PR #64, opened 2022-03-26
Use meilisearch instead of redisearch

razzeee:
Version of https://github.com/flathub/backend/pull/168 for the new repo


razzeee:
The CI error is caused by `docker compose up --no-build -d`
Especially the `no-build` part

PR #67, opened 2022-03-26
Change fields to correctly reflect, that we return the installs

razzeee:
New repo version of https://github.com/flathub/backend/pull/165 and https://github.com/flathub/frontend/pull/276

I'm still puzzled, why we think it shouldn't be `installs` in the frontend, as I find that much clearer from a user POV.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -104,7 +104,7 @@ def get_category(
 
     ids = apps.get_category(category)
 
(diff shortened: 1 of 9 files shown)


JakobDev:
@razzeee Is this API now safe for the shields.io badge?

razzeee:
yes

PR #68, opened 2022-03-27
Add some tasks for convenience

razzeee:
I couldn't figure out how to have docker tasks depend on each other and correctly wait for them to be started all the way. So this will need to be done by hand, otherwise it should be much nicer to use, if your using vscode now.

I also expect, that the tasks keep changing a bit in the start.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -40,8 +40,8 @@ If you want to run the smoketests locally, you can use the following commands:
 
 ```bash
    docker-compose up -d
-   docker exec backend_backend_1 pip3 install pytest
-   docker exec backend_backend_1 python3 -m pytest tests/main.py
+   docker exec backend-backend-1 pip3 install pytest
+   docker exec backend-backend-1 python3 -m pytest tests/main.py
 ```
 
 You might need to flush your redis database before running the tests. As it assumes that the database is empty.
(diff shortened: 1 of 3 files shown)


backend/README.md
```diff
@@ -40,8 +40,8 @@ If you want to run the smoketests locally, you can use the following commands:
 
 ```bash
    docker-compose up -d
-   docker exec backend_backend_1 pip3 install pytest
-   docker exec backend_backend_1 python3 -m pytest tests/main.py
+   docker exec backend-backend-1 pip3 install pytest
```
barthalion, line 43: It uses underscores for me with Docker.

barthalion: It would be probably more portable to replace it with `docker compose exec backend …`.

razzeee: I'm not sure why?
We're using it with `-` in the CI pipeline too and it will fail for me locally with underscore https://github.com/flathub/website/blob/a52252e8a6837ff77d8755eb9ba80818f0f79d1f/.github/workflows/backend_ci.yml#L80

barthalion: Huh, that's weird. Let's keep it as is then.

barthalion (approved):

PR #79, opened 2022-03-30
Add vcs-browser and contribute URL type

JakobDev:
These 2 URL types [have been added to AppStream](https://github.com/ximion/appstream/pull/392). This PR this Links to the new Website. They are already [supported by appstream-glib](https://github.com/hughsie/appstream-glib/pull/438), but not in a stable release yet, so it's currently not possible to upload a AppStream file with this new types to Flathub. This PR has not been tested yet. We may also need better Icons.

### DIFF
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -41,6 +41,8 @@
     "help": "Help",
     "contribute-translations": "Contribute Translations",
     "report-an-issue": "Report an Issue",
+    "vcs_browser": "Browse the source code",
+    "contribute": "Contribute to the App",
     "no-changelog-provided": "No changelog provided",
     "changes-in-version": "Changes in version {{version-number}}",
     "manual-install": "Manual Install",
--- a/frontend/src/components/application/AdditionalInfo.tsx
+++ b/frontend/src/components/application/AdditionalInfo.tsx
@@ -11,6 +11,8 @@ import {
   MdQuestionAnswer,
   MdTranslate,
   MdWeb,
+  MdCode,
+  MdEngineering,
 } from 'react-icons/md'
 import { BsHddFill, BsTextParagraph } from 'react-icons/bs'
 import { MdLaptop } from 'react-icons/md'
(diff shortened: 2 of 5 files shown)


frontend/src/components/application/AdditionalInfo.tsx
```diff
@@ -213,6 +213,42 @@ const AdditionalInfo = ({
...
         />
+      )}
+       {data.urls?.vcs_browser && (
+        <ListBox
+          appId={appId}
+          items={[
+            data.urls.vcs_browser
+              ? {
+                icon: <MdWeb />,
+                header: t('vcs_browser'),
```
razzeee, line 224: You'll need to add this and the `contribute` one to the english translation file.

frontend/src/components/application/AdditionalInfo.tsx
```diff
@@ -213,6 +213,42 @@ const AdditionalInfo = ({
...
           ]}
         />
+      )}
+       {data.urls?.vcs_browser && (
+        <ListBox
+          appId={appId}
+          items={[
+            data.urls.vcs_browser
+              ? {
+                icon: <MdWeb />,
```
razzeee, line 223: Maybe we can find some unique icons, that express this better for both.

JakobDev:
Both are now in the english translation file. I also found 2 icons that fit very well.

frontend/public/locales/en/common.json
```diff
@@ -41,6 +41,8 @@
     "help": "Help",
     "contribute-translations": "Contribute Translations",
     "report-an-issue": "Report an Issue",
+    "vcs_browser": "Browse the source code",
+    "contribute": "Contribute to the App",
```
bilelmoussaoui, line 45:
```suggestion
    "contribute": "Contribute to the application",
```
maybe?

razzeee: I actually think we would want these shorter and to match the others just `Contribute` would be better. And from reading the other PR, it sounded a bit more like it should be `How to contribute` if we want a long version.

JakobDev:
Succeeded by #259

PR #88, opened 2022-03-30
[Backend]Add AppStream translations

JakobDev:
This adds Support for AppStream translations. To use this, call /appstream/appid with the new lang parameter e.g. appstream/org.gimp.GIMP?lang=de. You will get the translated AppStream. This PR is currently WIP. Some fields are missing in the translation. I want to hear some opinions before going further.

### DIFF
--- a/backend/app/apps.py
+++ b/backend/app/apps.py
@@ -5,7 +5,7 @@ from . import db, utils
 
 
 def load_appstream():
-    apps = utils.appstream2dict("repo")
+    apps, apps_locale = utils.appstream2dict("repo")
 
     current_apps = {app[5:] for app in db.redis_conn.smembers("apps:index")}
     current_categories = db.redis_conn.smembers("categories:index")
@@ -46,6 +46,10 @@ def load_appstream():
                 for category in categories:
                     p.sadd("categories:index", category)
                     p.sadd(f"categories:{category}", redis_key)
+
+            for key, value in apps_locale.items():
+                if appid in value:
+                    p.set(f"apps_locale:{key}.{appid}", json.dumps(value[appid]))
 
         for appid in current_apps - set(apps):
             p.delete(
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -122,9 +122,15 @@ def list_appstream():
 
 
 @app.get("/appstream/{appid}", status_code=200)
-def get_appstream(appid: str, response: Response):
+def get_appstream(appid: str, response: Response, lang: str = None):
     if value := db.get_json_key(f"apps:{appid}"):
-        return value
+        if lang:
+            if translation := db.get_json_key(f"apps_locale:{lang}.{appid}"):
+                return value | translation
+            else:
+                return value
+        else:
+            return value
 
     response.status_code = 404
     return None
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -40,7 +40,7 @@ def add_translation(apps_locale: dict, language: str, appid: str, key: str, valu
     apps_locale[language][appid][key] = value
 
 
-def appstream2dict(reponame: str, language: str = None):
+def appstream2dict(reponame: str):
     if config.settings.appstream_repos is not None:
         appstream_path = os.path.join(
             config.settings.appstream_repos,
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -2,6 +2,7 @@ import gzip
 import hashlib
 import json
 import os
+import re
 
 import requests
 from lxml import etree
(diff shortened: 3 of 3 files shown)


razzeee:
AFAIK @barthalion is very against this, as long as the submission process isn't amended to handle translation better (?)

bilelmoussaoui:
Note that appstream has a tag `<language>` that can tell you how much of the application is translated in a specific language. Although there are two issues to note here:

- Whether the appstream file is marked as translatable in the upstream project
- Various apps don't have an appstream file upstream and it is only used in the Flathub repo sadly, which means there are no options for getting translations to it easily. 
- The language tag only work with `gettext`, so it won't work if you are translating it using a different mechanism.

JakobDev:
@bilelmoussaoui You missundertood. The <language> trag has nothing to do with this. YOu can translate AppStream using a xml:lang attribute.

Example:
```xml
<summary>This is a summary</summary>
<summary xml:lang="de">Dies ist eine kurze Beschreibung</summary>
```
In this case, the Summary is translated into german. You can do this with almost every tag. A real life exmaple would be the [AppStream of Kate](https://invent.kde.org/utilities/kate/-/blob/master/kate/data/org.kde.kate.appdata.xml).

If a AppStream file has no translation, that's not our problem. In this case, we simply show the English version. It's the Job of the Flathub website to support the AppStream Standard as good as possible, not to provide the AppStream data itself. Thats the Job of Upstream or the Maintainers.

@razzeee Did you mean how new Languages get added through Weblate?

bilelmoussaoui:
@JakobDev I am aware of that. What I mentioned is probably information to detect whether the appstream file is translated or not. Otherwise you will ship a mix of english/$other language in the same app page

JakobDev:
The solution is simple: If a Tag has a Translation, use it. If not, use the English version. It's the same as any other Software center does. The language tag only says how much the App itself is translated. It says nothing about the AppStream.

I personlay see no problem, if the Description and Summary is translated, but not the Changelog. It's the same behaviour that you find on other places on the Web e.g. the [AMO site of uBlock Origin](https://addons.mozilla.org/de/firefox/addon/ublock-origin/) (which is probably way more visited than a Apppage on Flathub) has a translated description but a English Changelog.

I see it simple: If the Upstream (Maintainers on Flathub, who adds a AppStream because Upstream has none, doesn't use translations) wants to show Users of a specific language a translated text, we should respect that. It's the choice of the Upstream how he want to present his App, not ours.

barthalion:
This is not going to happen without tightening appstream validation, including enforcing the list of required languages it should be available in, which I'm sure would cut off over half of the available applications on Flathub. So really, not going to happen in foreseable future at all.

JakobDev:
You don't need to force all languages that Flathub has in the AppStream. If a translation exists, it will be used. Otherwise the English version will be used. It is designed to work with existing AppStream Data without breaking anything.

ovari:
> I want to hear some opinions before going further.

Yes, please implement this feature.

This feature can help non-English speaking people more easily discover applications.

The screenshots at https://github.com/flathub/website/issues/611#issuecomment-1320946918 will then shown in Hungarian.

1. Please show the translated application name for translated applications like "Video Downloader", "Video Trimmer", "PDF Arranger", etc.
2. Please show the translated application summary for translated applications like "Octave" etc.
3. Please show the translated application description for translated applications too.
4. When searching, please search all the application names in all translations too. The Linux Mint operating system does this and is very helpful.

Thank you

heymisphere:
[apps.gnome.org](apps.gnome.org) has this feature, if translated summaries and names are present in metainfos, that website displays them

flipflop97:
I don't completely understand why using the existing appstream translations would be a problem, but that's not my decision to make. I do want to make an extra argument in favour of it though:

Right now the website itself can be translated, and has been translated in multiple languages. The app info however, is always shown in English. This means the website for languages other than English will always be bilingual (see an example below). Having translated appstream data will reduce this problem and will make many pages completely translated instead of partially, which means users won't have to switch between English and their native language all the time.

afbeelding

razzeee:
This is shipping now, thank you

razzeee:
That shouldn't be the case, are you sure the app provides translated captions?

razzeee:
Only thing in aware of that still is English on the app page is the release note

razzeee:
> > > That shouldn't be the case, are you sure the app provides translated captions?
> > 
> > 
> > Yes, the AppStream metadata does come with translated captions. However, after some more testing captions are not always in en only but sometimes. I am unable to figure out exactly when and when not, or why.
> 
> Furthermore, it turns out that if localized captions are displayed their order with regards to its attributed images/videos is messed up. 😵‍💫

Possible, was hard to test due to the redis app being broken and not knowing, which apps have translations. Will look into it.

razzeee:
I think that should all be fixed after https://github.com/flathub-infra/website/pull/3333

JakobDev:
It is also possible to have localized Screenshots (not only captions) now. They should also be supported.

razzeee:
> It is also possible to have localized Screenshots (not only captions) now. They should also be supported.

I can't even figure out how that should work, from looking at the appstream docs :/
Really not a fan of that website

JakobDev:
> I can't even figure out how that should work, from looking at the appstream docs :/

You can use `xml:lang` on the source link. See [this example](https://codeberg.org/JakobDev/jdReplace/src/commit/f3bb8532f179526cab939aea303a723e64bac511/deploy/page.codeberg.JakobDev.jdReplace.metainfo.xml#L18).

PR #89, opened 2022-03-30
(vending): Initial vending model and API

danielsilverstone-ct:
This initial vending model etc. adds the `/vending` URL namespace, and adds a status endpoint, an endpoint to start/continue onboarding, should the vendor need attention, and a dashboard link generator which we should use if the user wishes to log into their express dashboard.

This is only the start of vending, but I'm trying to keep PRs smaller to make them easier to review.  Once @ajr0d comes up with an approach for testing Stripe related stuff, we can add tests for this.


backend/alembic/versions/ff2ed634a1e9_initial_vending_models.py
```diff
@@ -0,0 +1,36 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ff2ed634a1e9'
+down_revision = '7589e5e508b9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('stripeexpressaccount',
```
razzeee, line 21: Hrm, having this non generic might adding an additional provider like paypal problematic. Still, we could just migrate the database then, so shouldn't be a blocker.

PR #107, opened 2022-04-05
Make URLs in Description clickable

JakobDev:
Some Apps e.g. Veloren have Links in their Description. I know, that the Description is not the right place for Links according to the AppStream Standard, but this doesn't stop Apps from doing this. This PR makes Links in Descriptions clickable.

### DIFF
--- a/frontend/src/components/application/Details.tsx
+++ b/frontend/src/components/application/Details.tsx
@@ -63,6 +63,14 @@ function categoryToSeoCategory(category) {
   }
 }
 
+function makeUrlsClickable(text) {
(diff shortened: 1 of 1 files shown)


nikodunk:
This would be very useful for forwarding users to the readme, as we need to do for VSCode as additional steps are required to escape the sandbox. https://beta.flathub.org/apps/details/com.visualstudio.code

razzeee:
Could be interesting to use https://linkify.js.org/docs/ for this

razzeee:
Closing in favor of the linked MR

PR #110, opened 2022-04-07
Darken main color for dark mode

razzeee:
Closes https://github.com/flathub/website/issues/109

### DIFF
--- a/frontend/pages/statistics.tsx
+++ b/frontend/pages/statistics.tsx
@@ -17,9 +17,11 @@ import {
 } from 'react-icons/md'
 import ListBox from '../src/components/application/ListBox'
 import { i18n, useTranslation } from 'next-i18next';
+import { useTheme } from 'next-themes';
 
 const Statistics = ({ stats }: { stats: Statistics }): JSX.Element => {
   const { t } = useTranslation()
+  const { resolvedTheme } = useTheme()
   let country_data: { country: string; value: number }[] = []
   if (stats.countries) {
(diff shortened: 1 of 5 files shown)


razzeee:
@jimmac

jimmac:
I'd love to get more involved and do reviews/contribute. I tried setting up the website locally, but couldn't make it work, sadly. Screenshots is the only way, I'm afraid.

razzeee:
Sorry I've missed your response in the issue somehow :(

But it would be interesting, why you couldn't make it work on your system, as I want the readme to be able to guide anyone through that.

razzeee:
A darker one

image

image

jimmac:
This works for the header 👍. 

I'd separate the buttons and give them more contrast vs the background (the old blue worked there ;)

razzeee:
Which button are you talking about the primary (`more` for e.g.) or the secondary `quick setup` or do you mean both?

image

Small change in the screenshot as the secondary button previously had both blue

jimmac:
> Which button are you talking about the primary (`more` for e.g.) or the secondary `quick setup` or do you mean both?

Both will benefit from stronger contrast. The original issue was about the large background spaces rather than individual controls. Having a single color property sounds nice and consistent, but perhaps deriving one from the other (value, opacity shift) allows to hit both objectives.

razzeee:
Let's merge this then and keep refining it? There are more places around the whole page, where this color change will show effects. Pretty hard to show, without actually using/deploying the page.

PR #113, opened 2022-04-07
Add purchase pages (front and backend)

flyingpimonster:
Add a backend endpoint for generating download tokens for apps. This API checks a database table, UserOwnedApp, to see if the user owns the given app ID (it also allows .Locale and .Debug suffixes). Populating the UserOwnedApp table will be part of the checkout flow, and is not part of this PR.

Add three pages on the frontend. The main one, `/purchase`, attempts to generate a download token for the requested app(s) and redirects based on the result: either to the login page, checkout page, or flathub-authenticator (which is running on localhost).

The current transaction "shopping cart" is stored in localStorage. If a new transaction is started, the old one will be replaced--I'm not sure if there's a good way around this. Once you log in, if there's a pending transaction, you'll be redirected to `/purchase` instead of your user page.

The checkout page is currently nearly empty. It has a `usePendingTransaction()` call so that in the future, it can access the `pendingTransaction.missingAppIDs` property to get the list of apps that need to be bought.

Depends on flatpak/flat-manager#53 for flat-manager to accept the download tokens created by this backend.


backend/app/config.py
```diff
@@ -38,6 +38,7 @@ class Settings(BaseSettings):
     stripe_secret_key: Optional[str] = None
     stripe_public_key: Optional[str] = None
     stripe_webhook_key: Optional[str] = None
+    flat_manager_secret: str = "c2VjcmV0"
```
razzeee, line 41: This is probably a placeholder? Might be nicer, to have the default be something, that will cause a failure, so you know, when you've set it correctly/at all.

frontend/public/locales/en/common.json
```diff
@@ -121,6 +121,12 @@
     "install": "Install",
     "screenshot": "Screenshot",
     "donate": "Donate",
+    "checkout": "Checkout",
+    "purchase-apps-title": "Purchase Apps",
+    "thank-you-for-your-purchase": "Thank you for your purchase!",
+    "safe-to-close-page": "It is safe to close this page.",
+    "app-install-error-try-again": "Could not install the app. Try going back to the app store and installing again.",
+    "checkout-apps": "Checkout",
```
razzeee, line 129: Different key, but same string as above (line 124)

backend/app/main.py
```diff
@@ -204,6 +212,58 @@ def get_summary(appid: str, response: Response):
...
+    if not login["state"].logged_in():
+        return JSONResponse({"detail": "not_logged_in"}, status_code=401)
+    user = login["user"]
+
+    def canon_app_id(app_id: str):
+        """
+        For .Locale, .Debug, etc. refs, we only check the base app ID. However, when we generate the token, we still
+        need to include the suffixed version.
+        """
+        auto_suffixes = ["Locale", "Debug"]
```
razzeee, line 228: Can't find any examples, but I'm wondering, if these can end up being in lowercase? Then this logic would fail.

backend/app/models.py
```diff
@@ -413,3 +413,49 @@ def by_transaction(cls, db, txn: Transaction):
...
+    def user_owns_app(db, user: FlathubUser, app_id: str):
+        return (
+            db.session.query(UserOwnedApp)
+            .filter_by(account=user.id, app_id=app_id)
+            .first()
+            is not None
+        )
+
+    @staticmethod
+    def all_by_user(db, user: FlathubUser):
```
razzeee, line 440: Maybe `all_owned_by_user` would be clearer?

backend/app/models.py
```diff
@@ -413,3 +413,49 @@ def by_transaction(cls, db, txn: Transaction):
...
+        return db.session.query(UserOwnedApp).filter_by(account=user.id)
+
+    @staticmethod
+    def delete_hash(hasher: utils.Hasher, db, user: FlathubUser):
```
razzeee, line 444: I don't think this is `delete`?

frontend/pages/purchase/finished.tsx
```diff
@@ -0,0 +1,27 @@
+import { GetStaticProps } from 'next'
+import { useTranslation } from 'next-i18next'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { NextSeo } from 'next-seo'
+import Main from '../../src/components/layout/Main'
+
+export default function Purchase() {
+  const { t } = useTranslation()
+
```
razzeee, line 9: Wondering if we should track this in matomo as an event, so that we can analyze this in there. We could also count in the database, but this might help show if somebody tries to exploit the api.

backend/app/main.py
bilelmoussaoui: They can't be as far as I know

backend/app/models.py
jameswestman: I'm not sure what you mean. It's added to TABLES_FOR_DELETE so we can remove a user's info when they delete their account. Although, maybe it's worth discussing if we really want to forget app ownership.

backend/app/config.py
jameswestman: If I set it to None here, is there a way to override it for development?

razzeee: There seems to be https://pydantic-docs.helpmanual.io/usage/settings/#environment-variable-names

But I don't think I'm a fan of that either, so let's keep it like it is.

frontend/pages/purchase/finished.tsx
razzeee: I would be looking for a suspicious amount of calls from the same user on the same/a bunch of apps. Not sure this was the correct place to do this, more a general idea. You will probably know better where to do this. I think it might also just be something, we might want an issue on to keep it in our head, without blocking you.

backend/app/models.py
razzeee: Nevermind, that code is basically already in place for each model. I don't think I understand what it does, but I'm fine with that.

backend/app/config.py
danielsilverstone-ct: I override the stripe stuff for development using envvars, you can see how in the docker-compose.yml - it's fairly easy.  But also for this kind of thing, just having a simple "this is a test key" or somesuch as the key will do I think.

backend/app/models.py
danielsilverstone-ct: This will always be invoked with the class name on the front, so `UserOwnedApp.all_by_user(db, login["user"])` or similar. As such, I don't think repeating `owned` is necessary.

danielsilverstone-ct (approved):
I can only speak from the PoV of the backend stuff; but that looks sane to me.

backend/app/config.py
jameswestman: This is "secret" in base64, same as the default in the flat-manager repo.

backend/app/models.py
razzeee: So this is more `mark_for_deletion` then actual `deletion`? That's probably why I was confused by the add.

razzeee (approved):

kymckay (approved):
Frontend-wise looks good to me, good find on the local storage hook!

Might be an idea to include a code comment directing to the source as could be useful for future devleopers looking at it too since it explains why some of the conditions are present (it's released under the [Unlicense](https://github.com/uidotdev/usehooks/blob/master/LICENSE) so no issue there). They have a `useAsync` function which I might grab too :smiley:

jameswestman:
Rebased. @kymckay Also added a comment with link.

razzeee:
Is this waiting for something or should we hit the button?

Issue #134, opened 2022-04-13
Improvements to payment elements and donation flow


kymckay:
A set of follow up tasks for #90 , ordered by priority:

- [x] Pagination to navigate a longer transaction history (requires some additional logic to correctly use the API)
- [x] Add confirmation to card deletion
- [ ] Back button in checkout flow (to return to card selection)
 - [x] Sensible navigation links from transaction pages (e.g. "Return to wallet" from transaction summary page, "Return to user page" from wallet)
 - [x] Code cleanup, I think we should start factoring the API interaction methods to some common place instead of putting them in component source files (there are definitely a few similar functions in here). Perhaps an actions directory with `login.ts`, `payment.ts` etc.

In the longer term, these are also desirable, but can perhaps be broken out into seperate issues as and when needed:

- [x]  A reusable component for pages where login is required that sends the user back to where they were after login (e.g. visit donate page, directed to login, directed back to donate page).
- [ ] Move towards better integration of wallet elements on userpage

PR #147, opened 2022-04-20
Add page traversal to transaction history

kymckay:
- Can't use arbitrary page jumping since retrieval API works sequentially, so user can use "next" and "previous" page buttons (or jump to start)
- Selecting sorting complicates the logic and has been left out for now

Part of #134

### DIFF
--- a/frontend/src/components/payment/transactions/TransactionHistory.tsx
+++ b/frontend/src/components/payment/transactions/TransactionHistory.tsx
@@ -1,10 +1,13 @@
 import { useTranslation } from 'next-i18next'
 import { FunctionComponent, useEffect, useState } from 'react'
+import { toast } from 'react-toastify'
 import { useUserContext } from '../../../context/user-info'
 import { TRANSACTIONS_URL } from '../../../env'
 import { Transaction } from '../../../types/Payment'
+import Button from '../../Button'
 import Spinner from '../../Spinner'
 import TransactionList from './TransactionList'
+import styles from './TransactionHistory.module.scss'
 
 async function getTransactions(
   sort: string = 'recent',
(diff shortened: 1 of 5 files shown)


razzeee:
This might be easy to improve for anyone that wants to write python code (@danielsilverstone-ct for e.g.) as it seems, that the paging api doesn't have to be cursor based

frontend/src/components/payment/transactions/TransactionHistory.tsx
```diff
@@ -65,10 +93,44 @@ const TransactionHistory: FunctionComponent = () => {
...
+              page * perPage + perPage
+            )}
+          />
+          <div className={styles.controlPage}>
+            <Button
+              type='secondary'
+              onClick={() => setPage(0)}
+              disabled={page === 0}
+            >
+              {t('first-page')}
```
razzeee, line 109: Do we really need first page? Wouldn't expect that to be used often and throws of the balance of the buttons?

razzeee (approved):
Minor nitpick, seems fine otherwise

razzeee:
We could also use the Material Icons instead of text 
image

(without any additional styling/sizing)

frontend/src/components/payment/transactions/TransactionHistory.tsx
kymckay: Yeah let's rip it out, I was on the fence too

kymckay:
Good shout about the material icons, I like those more, but probably would still want the text to prompt screen readers, etc.

I won't be working tomorrow so if you want to push and merge before Friday feel free to go ahead :smiley:

PR #148, opened 2022-04-20
Add LoginGuard convenience wrapper component

kymckay:
- An easy way to only render a page's content when we know the user is logged in. Otherwise redirecting them to login.
- Simplifies some code by factoring this pattern out to a new component.
- For user QoL it would be handy if this redirected back to the same page following login, but need to think further about how best to handle that (probably local storage, but would need to guard against unexpected redirection if user doesn't login then later does).


razzeee:
Looks good to me, would be a nice case for some tests.

PR #160, opened 2022-04-22
Add vending link to userpage

kymckay:
- Only shown if user has 1 or more apps
- Uses the vending API to check status of signed-in user's external vending account
- Presents either onboarding button or dashboard link as needed
- Alerts user to update their account if attention is needed

A little unsure how best to present these elements , but have opted for 

- A subsection (not 100% happy with my chosen heading) at the bottom of the user page's top element (below linked accounts)
- A button to begin onboarding - because clicking it initiates an action to fetch an onboarding link and sends browser there (user is redirected back upon completion). This link can't be pre-fetched otherwise Express accounts will be created without user action.
- A plain link to the Stripe Express dashboard after onboarding (so that it can open in a new window), with conditional red text shown underneath when the account needs attention from the user (as reported by the API)

Here's how it looks on the user page before (and during incomplete) onboarding:
Screenshot from 2022-04-25 15-18-29

After onboarding but when my test details fail to validate:

Screenshot from 2022-04-25 15-36-41

### DIFF
--- a/frontend/src/asyncs/vending.ts
+++ b/frontend/src/asyncs/vending.ts
@@ -1,0 +1,80 @@
+import {
+  VENDING_DASHBOARD_URL,
+  VENDING_ONBOARDING_URL,
+  VENDING_STATUS_URL,
+} from "../env"
+import { VendingRedirect, VendingStatus } from "../types/Vending"
+
+// API responds with this status code if onboarding has not begun
+const STATUS_NEW = 201
+
+const DEFAULT_STATUS: VendingStatus = {
+  can_take_payments: false,
+  details_submitted: false,
+  needs_attention: false,
+}
+
+/**
+ * Retrieve the vending status of the logged in user.
+ * @returns The vending status object
+ */
+export async function getVendingStatus(): Promise<VendingStatus> {
+  let res: Response
+  try {
+    res = await fetch(VENDING_STATUS_URL, { credentials: "include" })
+  } catch {
+    throw "failed-to-load-refresh"
+  }
+
+  if (res.ok) {
+    if (res.status === STATUS_NEW) {
+      return DEFAULT_STATUS
+    }
+
+    const data: VendingStatus = await res.json()
+    return data
+  } else {
+    throw "failed-to-load-refresh"
+  }
+}
+
+export async function getDashboardLink(): Promise<string> {
+  let res: Response
+  try {
+    res = await fetch(VENDING_DASHBOARD_URL, { credentials: "include" })
+  } catch {
+    throw "failed-to-load-refresh"
+  }
+
+  if (res.ok) {
+    const data: VendingRedirect = await res.json()
+    return data.target_url
+  } else {
+    throw "failed-to-load-refresh"
+  }
+}
+
+export async function getOnboardingLink(): Promise<string> {
+  let res: Response
+  try {
+    res = await fetch(VENDING_ONBOARDING_URL, {
+      method: "POST",
+      credentials: "include",
+      headers: {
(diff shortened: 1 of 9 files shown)


kymckay:
Having experimented with the onboarding and API responses, I have an assumption I want to confirm before this is ready to make sure we're reliably distinguishing between:
- "has begun onboarding, but not finished"
- "account requires attention at some point in future after onboarding"

I definitely know dashboard links can only be created once `can_take_payment` becomes true. However, this does not mean the opposite holds which is the only condition I can use to determine if the user has started onboarding and not finished (thus, to retrieve an onboarding link instead of a dashboard link).

I have a feeling it's not an adequate condition as it seems sensible to me that accounts could lose their "transfer" capability (what the backend actually checks).

@danielsilverstone-ct likely of interest to you

danielsilverstone-ct:
I think we can create a dashboard link once we've reached can_take_payment at least once; nominally if the user is maintaining their stripe account, then that'll never "go false", but the needs-attention flag should cause us to show a "your account needs attention, click here" type link 

Check the stripe docs for express accounts; if you think we need different flags visible to us; then we can adjust the model in our cache of the account status

kymckay:
https://stripe.com/docs/connect/express-accounts#handle-users-not-completed-onboarding

> You can check the state of the details_submitted parameter on their account to see if they’ve completed the onboarding process.

This seems like what we need to correctly determine whether to show the onboarding link or the dashboard link since both other booleans can be in either state after onboarding. I've gone ahead and included the backend change here for that.

frontend/src/components/user/VendingLink.tsx
```diff
@@ -0,0 +1,169 @@
...
+import styles from "./VendingLink.module.scss"
+
+// API responds with this status code if vending has not begun
+const STATUS_NEW = 201
+
+/**
+ * Retrieve the vending status of the logged in user.
+ * @returns The vending status object, or nothing if onbording has not started.
+ */
+async function getStatus() {
```
razzeee, line 22: Move to async folder

frontend/src/components/user/VendingLink.tsx
```diff
@@ -0,0 +1,169 @@
...
+  let res: Response
+  try {
+    res = await fetch(VENDING_STATUS_URL, { credentials: "include" })
+  } catch {
+    throw "failed-to-load-refresh"
+  }
+
+  if (res.ok) {
+    if (res.status === STATUS_NEW) {
+      return
```
razzeee, line 32: Not a fan of functions that return vastly different results. At least annotate it correctly with `Promise<VendingStatus | undefined>` as typescript doesn't seem to pick up on that correctly (with the current config)

razzeee:
Can this be tested with the default config? I only ever get an error with that.

razzeee (approved):
Some smaller nitpicks, nothing major

kymckay:
> Can this be tested with the default config? I only ever get an error with that.

Currently no, when the stripe keys aren't present there's no fake version of these routes in the backend

frontend/src/asyncs/vending.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const DEFAULT_STATUS: VendingStatus = {
+  can_take_payments: false,
+  details_submitted: false,
+  needs_attention: false,
+}
+
+/**
+ * Retrieve the vending status of the logged in user.
+ * @returns The vending status object, or nothing if onbording has not started.
```
razzeee, line 19: Seems outdated now

kymckay: Good catch :+1:

PR #161, opened 2022-04-22
Use useAsync hook to simplify account deletion

kymckay:
- Introduces a handy useAsync hook
- Simplify state handling in account deletion component using the new hook instead

Adding this hook is a first step towards some API-based code clean up. Going to convert more to use it when I start centralising these API functions, but submitting this small patch now before I sign off for the week.


frontend/src/components/user/DeleteButton.tsx
```diff
@@ -4,103 +4,78 @@ import { toast } from 'react-toastify'
...
+
+  if (status === 'pending') {
     return <Spinner size={30} />
   }
 
   if (token) {
-    // Callbacks must clear the token to hide the dialog
     return (
       <ConfirmDialog
         prompt='Delete your account?'
```
razzeee, line 68: Are these missing translation?

frontend/src/components/user/DeleteButton.tsx
```diff
@@ -4,103 +4,78 @@ import { toast } from 'react-toastify'
...
         entry='I wish to delete my account'
```
razzeee, line 69: Might be weird to translate this, not sure I have a better idea.

PR #172, opened 2022-04-28
Use prettier and format accordingly

razzeee:
It seems like we had prettier rules, but they were not enforced.

Not sure about the translation files, we might need to exclude them, if they keep on changing indention.


razzeee:
@barthalion added, please double check

frontend/.prettierrc.json
```diff
@@ -1,6 +1,4 @@
 {
-  "trailingComma": "es5",
-  "semi": false,
-  "singleQuote": true,
-  "jsxSingleQuote": true
```
kymckay, line 5: Any reasoning for switching from single quotes?

kymckay (commented):
I notice we recommend the Prettier extension to those using VSCode, is it also worth adding the project-level config to set it as the default formatter where relevant?

frontend/.prettierrc.json
razzeee: Good question, I usually like to leave this as default as possible. I would usually only set trailingComma to `all` and nothing else. Decided to keep semi `false`, as it seems pretty widespread in the solution.

razzeee:
> I notice we recommend the Prettier extension to those using VSCode, is it also worth adding the project-level config to set it as the default formatter where relevant?

I'm not sure which config that is or how that would look. I would think we have most things configured after this PR.

kymckay:
> I'm not sure which config that is or how that would look.

Under the `.vscode` directory a `.settings` can set workspace settings like:

```
{
  "editor.formatOnSave": true,
  "[javascript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "[typescriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  }
}
```

It would be somewhat enforcing the use of that extension to any VSCode users working on this repo which may not be desirable though.

kymckay (approved):
lgtm

PR #175, opened 2022-04-29
Move API handling to common location in codebase

kymckay:
- Collect asynchronous API methods under a new `asyncs` directory  (where any other future async actions can go too) so that they're not dispersed between various component files.
- I'm also adding a base interface for API responses to give a type to the common JSON returned by endpoints (which can be extended for more specific responses).

This is purely a development QoL improvement, any other code changes are just formatting done by Prettier. I left the `fetchers.ts` file untouched since those async methods are all called statically and so don't interact with component state, etc.


frontend/src/types/API.ts
```diff
@@ -0,0 +1,4 @@
+export interface APIResponse {
+  status: "ok" | "error"
```
razzeee, line 2: Am I correct, that with the current code, this will never be "error"? 

Two things I personally like/do:
- I usually have aditional "notAsked"/"loading" states, which I'm not sure if they make sense for us here. See https://package.elm-lang.org/packages/krisajenkins/remotedata/latest/RemoteData#RemoteData for some previous work in a different ecosystem
- I'm wondering if we can type the response differently between success/error/etc so that it only has the correct fields.

kymckay: It can be error, we're just not often having to check for it because we send fetch requests when appropriate (only when logged in for example).

I don't think we can type them differently because (it's perhaps a bit misleading) this type is being driven by the current form of backend **successful** responses so there's no way to distinguish them before checking for these attributes.

However, I have just noticed that some use an `error` attribute where others use `message`. I'll revisit this and perhaps homogenise those in the backend.

kymckay:
I removed the base API interface from this PR in the interest of getting the code reshuffling in. Will need to revisit it and homogenise the backend responses at the same time.

frontend/src/types/API.ts
kymckay: This prompted me to re-look at the backend and realise I'd somehow got it into my head that the response with `status: "error"` could occur when the actual HTTP status was OK. In that case yeah this can be revisited properly with separate types :+1:

PR #178, opened 2022-04-29
Add confirmation to card deletion

kymckay:
Bit more awkward than expected since the state of the parent component needs to be updated on success. I originally opted to use a reducer, but this state is a simple array so went with a callback to remove cards in the end and reduce coupling slightly.

Clicking the remove saved card button now prompts to confirm or cancel (with tooltips shown on hover and aira-labels for sceeen readers):

Screenshot from 2022-05-04 11-31-56


kymckay:
Would look to get #175 in first and I'll rebase this to update the import paths

PR #184, opened 2022-05-01
Show addons on the detail page

### LINKED ISSUE #603: Display add-ons for apps
It would be great if there were also application add-ons displayed on Flathub.org. See for example how add-on section looks like in GNOME Software (on the details page of Steam):

Add-ons

razzeee:
This is a port of https://github.com/flathub/frontend/pull/128/files and https://github.com/flathub/backend/pull/87/files to the new repo.

It needs to be reworked. Especially on the frontend, as the way we display addons doesn't scale. We can probably use the collapse behavior we have now to make it nicer, but there are a few more things to do.

Closes #603


razzeee:
Even if I create a custom flatpakref by hand, it's no good. If I try to load it via g-s it seems to error out, probably as addons don't have their own pages.

PR #194, opened 2022-05-04
Move Login to header

ajr0d:
Closes https://github.com/flathub/website/issues/135.

Open to any suggestions or changes.


frontend/public/locales/en/common.json
```diff
@@ -54,7 +53,7 @@
     "network-error-try-again": "There was a network error. Try again.",
     "error-already-logged-in": "The selected account is already logged in.",
     "failed-to-load-refresh": "The resource failed to load. Try refreshing.",
-    "developer-login": "Developer Login",
+    "developer-login": "Login",
```
razzeee, line 56: I think this key needs to change too

frontend/src/components/login/Status.module.scss
```diff
@@ -4,15 +4,14 @@
   .user {
     display: flex;
     align-items: center;
-    background-color: rgba(255,255,255, 0.2);
     border-radius: var(--border-radius);
-    padding: 5px;
-    column-gap: 10px;
+    padding: 4px;
+    column-gap: 5px;
```
razzeee, line 9: Try to do spacing that's devisable by 2 or better 4 please :)

razzeee:
Feel free to squash this down to one commit, when your rebasing.

Here's a list of probems with the current implementation:
- Image should probably be on the right of the name
- Button is heigher then the header itself, ideally it would be of the same height as the ones next to it
- My name gets split into two rows kinda connected to the previous one
- We might need to ellipse names at a certain length
- displays completly wrong in mobile mode (collapsed header)

ajr0d:
> * Image should probably be on the right of the name

Done.

> * My name gets split into two rows kinda connected to the previous one

I think I've added a fix for this, but you might need to check on your end.

> * We might need to ellipse names at a certain length

Also done, but I think `overflow: hidden` may of made this rare from actual happening.

> *   displays completly wrong in mobile mode (collapsed header)

I think this is fixed now.

> * Button is heigher then the header itself, ideally it would be of the same height as the ones next to it

I'm guessing you mean when you highlight/hover over the button and the colour change is noticeably bigger than the other links? - If so, I'm still looking at this but it could be an issue while the profile image is active.

I will obviously tidy up the files and the commits once everyone is happy with this, and we take it out of draft status.

razzeee:
> I will obviously tidy up the files and the commits once everyone is happy with this, and we take it out of draft status.

I usually don't look at stuff that's not rebased, as that rebase will change the code and therefore the review :)

razzeee:
I've rebased and squashed this.

PR #196, opened 2022-05-04
Add convenient navigation links to checkout pages

kymckay:
A small "Go to X" type of link at the top of the page content for easy return to where a user came from (or to a related page)

Screenshot from 2022-05-04 14-07-41

Part of #134

### DIFF
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -202,5 +202,6 @@
   "vending-onboard": "Create a Stripe Express account",
   "requires-attention": "Requires attention",
   "cancel": "Cancel",
-  "delete": "Delete"
+  "delete": "Delete",
+  "go-to-page": "Go to {{page}}"
 }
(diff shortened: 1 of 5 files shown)


frontend/public/locales/en/common.json
```diff
@@ -202,5 +202,6 @@
   "vending-onboard": "Create a Stripe Express account",
   "requires-attention": "Requires attention",
   "cancel": "Cancel",
-  "delete": "Delete"
+  "delete": "Delete",
+  "go-to-page": "Go to {{page}}"
```
razzeee, line 206: I'm torn on this, I kinda think doing this in a breadcrumb style might be easier to handle, but not super sure I want to push that onto you.

kymckay: From a UX perspective that sounds like the better solution, unsure if it fits with our work packages but perhaps something I'd look at in my free time if nobody else is inclined. Say we push it out to a new issue for now?

PR #198, opened 2022-05-04
Direct user back to where they were upon login

kymckay:
Since login directs the user away from the site, we need to use local
storage to determine where to send them on return.

However, by sending them to the login page there's no guarantee that
they will go through with login. We don't want to pre-emptively set a
value which will later falsely be used if they return to login.

If that reasoning is unclear, also consider the scenario where a user
has two login windows open as a result of trying to visit different
protected pages. Whichever window they proceed with should direct them
back to where they were in that window, not the other window.

Using a URL query string means the local storage value can be set only
immediately before redirect away from the site, when we know they are
performing login. It can then be used from local storage on return.

If a value does remain in local storage, we prefer the query string
at time of redirect which will overwrite it.

### DIFF
--- a/frontend/pages/login/[service].tsx
+++ b/frontend/pages/login/[service].tsx
@@ -11,6 +11,7 @@ import { useUserContext, useUserDispatch } from "../../src/context/user-info"
 import { fetchLoginProviders } from "../../src/fetchers"
 import { useLocalStorage } from "../../src/hooks/useLocalStorage"
 import { usePendingTransaction } from "../../src/hooks/usePendingTransaction"
+import { isInternalRedirect } from "../../src/utils/security"
 
 export default function AuthReturnPage({ services }) {
   // Must access query params to POST to backend for oauth verification
@@ -32,8 +33,15 @@ export default function AuthReturnPage({ services }) {
       if (pendingTransaction) {
         router.push("/purchase")
       } else if (returnTo) {
-        router.push(decodeURIComponent(returnTo))
+        const redirect = decodeURIComponent(returnTo)
         setReturnTo(null)
+
+        // Must validate the redirect to prevent open redirect and code execution
+        if (isInternalRedirect(redirect)) {
+          router.push(redirect)
+        } else {
+          router.push("/userpage")
+        }
       } else {
         router.push("/userpage")
       }
--- a/frontend/pages/login/[service].tsx
+++ b/frontend/pages/login/[service].tsx
@@ -9,6 +9,7 @@ import { login } from "../../src/asyncs/login"
 import Spinner from "../../src/components/Spinner"
 import { useUserContext, useUserDispatch } from "../../src/context/user-info"
 import { fetchLoginProviders } from "../../src/fetchers"
+import { useLocalStorage } from "../../src/hooks/useLocalStorage"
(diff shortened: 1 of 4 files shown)


frontend/pages/login/[service].tsx
```diff
@@ -23,12 +24,16 @@ export default function AuthReturnPage({ services }) {
...
+  const [returnTo, setReturnTo] = useLocalStorage<string>("returnTo", "")
 
   useEffect(() => {
-    // Redirect to userpage once logged in. Or, if there's a pending transaction, redirect to the purchase page.
+    // Redirect back to where user was, or userpage for unprompted login
     if (user.info && !user.loading) {
       if (pendingTransaction) {
         router.push("/purchase")
+      } else if (returnTo) {
+        router.push(`/${returnTo.replace(".", "/")}`)
```
Comment (sorin-davidoi) [inline frontend/pages/login/[service].tsx:35]: I think this introduces the following security issues:
 - open redirect vulnerability - I can send someone a link to https://beta.flathub.org/login?returnTo=https:..evil.com and this code will redirect them to https://evil.com (which might impersonate the Flathub payment page, for example). See https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html.
 - remote code execution - I can send someone a link to `https://beta.flathub.org/login?returnTo=javascript:alert('Oh no!')` and this code will execute `alert('Oh no!')` in the user's browser.

We should probably validate that we are redirecting them inside the same application. The most straightforward way would probably be to ensure that `returnTo` is always absolute and validate the host:

```javascript
try {
   const parsedReturnTo = new URL(returnTo);

   if (parsedReturnTo.host === location.host) {
      // Safe to redirect
   }
} catch (e) {
  // Show error
}
```

Comment (sorin-davidoi) [reply frontend/pages/login/[service].tsx:35]: Ah, didn't notice the `/` at the beginning (shouldn't review code after midnight :smiling_face_with_tear:). That would prevent the two issues I've described, but resting the security of this code on that one character is probably not ideal (I can easily imagine someone removing it in the future, when the requirements change or during some refactoring, without realizing the security implications).

Comment (kymckay) [reply frontend/pages/login/[service].tsx:35]: Thanks for flagging this up :smiley:

I agree let's not rely on that `/`, I didn't fully think through the security concerns and intuitively landed on a simple solution which implicitly sweeps them aside. I'll rework this to include more explicit defensive code like you suggest :+1:

frontend/src/components/login/LoginGuard.tsx
```diff
@@ -18,7 +17,9 @@ const LoginGuard: FunctionComponent = ({ children }) => {
   // Content unsuitable if not logged in, send user to login page
   useEffect(() => {
     if (!user.info && !user.loading) {
-      router.push("/login")
+      router.push(
+        `/login?returnTo=${router.asPath.substring(1).replaceAll("/", ".")}`,
```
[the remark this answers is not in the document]
kymckay: Quick and dirty solution I forgot to replace. Will switch to `encodeURIComponent` and `decodeURIComponent` to handle the reserved characters more robustly.

razzeee (approved):

razzeee:
LGTM

PR #201, opened 2022-05-05
feat(frontend): add content-security-policy

sorin-davidoi:
Infer the current rules from https://beta.flathub.org and inscribe them in the policy. From my testing using Laboratory (Content Security Policy / CSP Toolkit) everything seems to work just fine.

The only caveat is that one inline script is blocked (`Content Security Policy: The page's settings blocked the loading of a resource at data:text/javascript;base64,IWZ1bmN0aW9u… ("script-src").`). This script is injected by the `next-themes` package, and the only way to make it run is to add `data:` to `script-src`, which is a trade-off that I don't think it's worth it, especially since the theme switching seems to work just fine without this script. Note that in the next version of the package we should be able to add a hash exception for this script: https://github.com/pacocoursey/next-themes/issues/106, which is a much better trade-off.

Note that I haven't tested being authenticated and anything payment related, so a few adjustments might be necessary.

The CSP string might not be very readable, but it's easy to test by copy-pasting it into Laboratory (Content Security Policy / CSP Toolkit) and similar tools, so I've left it like that (instead of storing it into an object and stringifying it later).

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -1,6 +1,17 @@
 const { PHASE_PRODUCTION_SERVER } = require("next/constants")
 const { i18n } = require("./next-i18next.config")
 
+const CONTENT_SECURITY_POLICY = `
+  default-src 'none';
+  script-src 'self' https://webstats.gnome.org;
+  style-src 'self' 'unsafe-inline' https://dl.flathub.org;
+  font-src 'self' https://dl.flathub.org;
+  connect-src 'self' https://flathub.org https://webstats.gnome.org;
+  img-src 'self' https://dl.flathub.org https://webstats.gnome.org data:;
+`
+  .replace(/\s{2,}/g, " ")
+  .trim()
+
 module.exports = (phase) => ({
   i18n,
   images: {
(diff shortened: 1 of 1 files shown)


razzeee:
Any opinion(s) on the approach in https://github.com/flathub/website/pull/200 ?

> The CSP string might not be very readable, but it's easy to test by copy-pasting it into Laboratory (Content Security Policy / CSP Toolkit) and similar tools, so I've left it like that (instead of storing it into an object and stringifying it later).

The next.js docs break the string by lines and replace linebreaks with spaces https://nextjs.org/docs/advanced-features/security-headers#content-security-policy

sorin-davidoi:
> Any opinion(s) on the approach in https://github.com/flathub/website/pull/200 ?

Ah, you got there first! Looks good to me, though I would personally not allow `data:` in `script-src`.

> The next.js docs break the string by lines and replace linebreaks with spaces

Nice, that would be increase readability while allowing easy debugging.

razzeee:
I'm very new to that, so that's what the mozilla generator gave me. I actually assumed `data:` was the same as just passing an url. Guess I need time to actually read the docs.

sorin-davidoi:
Looks like `phase === PHASE_PRODUCTION_SERVER` doesn't behave like I've expected, so the CSP is not applied on beta.flathub.org. Will take a look over the coming days.

PR #202, opened 2022-05-05
Add typing to API async methods

kymckay:
- Typing for API response JSON objects.
- Return types to asyncs to enforce correct and consistent returns.
- Fix some frontend error handling being unreachable.
- Fix a bad status code in the backend upon user deletion request error.

### DIFF
--- a/frontend/src/types/API.ts
+++ b/frontend/src/types/API.ts
@@ -1,0 +1,8 @@
+export interface APIResponseOk {
+  status: "ok"
+}
+
+export interface APIResponseError {
+  status: "error"
+  error: string
+}
--- a/frontend/src/types/Payment.ts
+++ b/frontend/src/types/Payment.ts
@@ -1,6 +1,7 @@
+import { APIResponseOk } from "./API"
+
 // Corresponds to `/wallet/walletinfo`
-export interface WalletInfo {
-  status: string
+export interface WalletInfo extends APIResponseOk {
   cards: PaymentCard[]
   account?: object
 }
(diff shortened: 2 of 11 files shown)


frontend/src/types/Payment.ts
```diff
@@ -15,6 +16,9 @@ export interface PaymentCard {
 }
 
 // Corresponds to objects from `/wallet/transactions`
+type TransactionKind = "donation" | "purchase"
+type TransactionStatus = "new" | "pending" | "retry" | "success" | "cancelled"
+
 export interface Transaction {
```
razzeee, line 22: Seems to only be used internally, so don't `export` it

frontend/src/types/Payment.ts
```diff
@@ -1,6 +1,7 @@
+import { APIResponseOk } from "./API"
```
razzeee, line 1: Any reason to not do the extends for `TransactionDetailed` too?

frontend/src/types/API.ts
```diff
@@ -0,0 +1,8 @@
+export interface APIResponseOk {
```
razzeee, line 1: I'm not really happy with this, maybe we can just generate everything from the openAPI/swagger definition, that the backend offers (I was really for graphQL, but that ship had already sailed, when I started working on this)? Then we should have correct types and hopefully also correct error types.

Have been looking at this for e.g. https://github.com/YBogomolov/fetcher-ts

razzeee:
Rebased this for you, as it was conflicting, will merge as soon as CI is green.

PR #237, opened 2022-05-11
feat(backend): add GNOME login support

kymckay:
Pretty simple as just another form of GitLab login.

Not familiar with SQLAlchemy at all, so looks like quite a lot of repeated code to me, but as far as I can tell just a necessary evil since we need unique tables for these.

Frontend SVG sourced from: https://foundation.gnome.org/logo-and-trademarks/ . Text stripped out (with permission) to make it square the same as other login logos.


ajr0d:
Looks good to me.

backend/app/logins.py
```diff
@@ -420,6 +453,70 @@ def gitlab_postlogin(tokens, account):
...
+    }
+    ```
+
+    This endpoint will either return an error, if something was wrong in the
+    backend state machines; or it will return a success code with an indication
+    of whether or not the login sequence completed OK.
+    """
+
+    def gnome_userdata(tokens):
+        gh = Gitlab("https://gitlab.gnome.org", oauth_token=tokens["access_token"])
```
razzeee, line 488: The `gh` is probably a c&p rror, but not that important

razzeee (approved):
Very small nitpick, feel free to ignore.

backend/app/config.py
```diff
@@ -30,6 +30,13 @@ class Settings(BaseSettings):
         "fe364dd63c43ba8ebab5bd488df5946adfcc14366edf81c3814f6a021c61e953"
     )
     gitlab_return_url: str = "http://localhost:3000/login/gitlab"
+    gnome_client_id: str = (
+        "914fb36557a5b8403f5f4d4085b58119b505aad07c56505ed14362a388aae6a8"
+    )
+    gnome_client_secret: str = (
+        "9122d58ed56789d80db5f1ed0ec6bad5d9dae612aa9d458d89ee475e01151c43"
+    )
+    gnome_return_url: str = "http://localhost:3000/login/gnome"
```
bilelmoussaoui, line 39: Is this expected?

kymckay: It is, these are the local development settings (also why the secrets are published here)

bilelmoussaoui: Ah I see, I suppose the return url has to be fixed on gitlab.gnome side? because it returns to localhost currently (from beta.flathub.org)

razzeee: I think barth needs to find time to setup the env vars

PR #243, opened 2022-05-11
feat(frontend): add referrer policy header

sorin-davidoi:
From https://infosec.mozilla.org/guidelines/web_security#referrer-policy:

> `strict-origin-when-cross-origin` send full referrer on same origin, URL without the path on foreign origin

This will prevent accidentally leaking user information from the URL to 3rd parties (e.g. if we ever create a path like `/user/:id`, 3rd parties shouldn't get the id in the header, but only `https://beta.flathub.org`).

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -49,6 +49,10 @@ module.exports = (phase) => ({
             value: "nosniff",
           },
           {
+            key: 'Referrer-Policy',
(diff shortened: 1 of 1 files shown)


razzeee:
At least our linter actually forces us to use `rel="noreferrer"` on `_blank` links at least. Still a good thing to have, especially, if someone doesen't even things via an `a` or blank

PR #244, opened 2022-05-11
Fix some linter warnings

razzeee:
This seems to break the avatar picture for me, any idea why that might be @kymckay


razzeee:
Oh it doesn't get lost, it only get's shrunk down, when there is no place

kymckay (approved):
Not sure I follow the above comments, but assuming that's resolved the avatar behaves as expected for me across screen widths (disappearing when header buttons become a menu instead)

PR #245, opened 2022-05-12
feat(frontend): add T&C's agreement to checkout

### LINKED ISSUE #240: Add positive confirmation of T&Cs at purchase stage
We need a check box when checking out that confirms that people have read the T&Cs, and that it's a digital download so there's no refunds.

> If you want to download something within 14 days of buying it, you will have to give your consent to waive the 14-day cooling-off period.
> If you don’t give your consent, the 14-day cooling-off period still applies, but you won’t be able to download your digital content until this period has ended.

I posit that we don't want the hassle of the latter, so we should just do the former.

kymckay:
Since payment can occur at the saved card selection stage or the new
card input stage, it's easiest to collect user T&C consent at the start
of checkout.

Two points we should clarify before merging:

- [ ] Should this be translated? Or do we want to ensure the language stays precise?
- [ ] Is there any specific language we should use in collecting consent here? I went with something that seemed reasonable, but we should make sure it's sound before merging.

closes #240


kymckay:
One thing to note: This won't be testable in staging until we actually generate purchase transactions (at the moment the backend only handles donations to Flathub itself which won't show this confirmation stage of checkout).

Here's a screenshot of me forcing it to show up in development:

Screenshot from 2022-05-12 14-38-51

razzeee:
@neilmcgovern  @ramcq

razzeee:
This conflicts and I would like to hear from neil if this is what we need

Issue #241, opened 2022-05-11
400 on log in attempt if the flow is interrupted


q30a:
**Steps to reproduce** (use private mode to start with a clean slate):
1. Go to https://beta.flathub.org/login
2. Click one of available options (e.g. GitHub)
3. You will be redirected to selected auth provider.
4. Go back one page and land on https://beta.flathub.org/login again.
5. Try to click other auth providers.

**Current behavior:**
400 on every attempt for providers other than in step 2.
`{
    "state": "error",
    "error": "User is mid-login-flow elsewhere"
}`
**Expected behavior:**
User can select auth providers freely.

q30a:
`{"state":"error","error":"User is mid-login-flow elsewhere"}`

kymckay:
Daniel and I discussed the ideal solution to this in a sync call one week. Believe we landed on the backend clearing out the in-progress flow and initiating the new flow (which requires some wrangling of the way the flow tokens are managed).

razzeee:
Possibly fixed by https://github.com/flathub/website/pull/981

rokejulianlockhart:
Indeed, @razzeee, it has been remediated for me.

PR #257, opened 2022-05-14
Change login flow handling to prevent 400 errors

ajr0d:
Spun up a quick fix (attempt) at what was pointed out on #241. I'm second guessing the fix however, as I'm not sure if and when the `else` statement would happen. Any improvements, please let me know and I'll add them.

### DIFF
--- a/backend/app/logins.py
+++ b/backend/app/logins.py
@@ -240,12 +240,13 @@ def start_oauth_flow(
 
     Examples of oauth flows include Github and Gitlab.
     """
-    if login["state"].logging_in():
+    while login["state"].logging_in():
+        login["method"] = method
         # Already logging in to something
         if login["method"] == method:
             # Already logging into correct method, so assume something went squiffy
             # and send them back with the same in-progress login
-            pass
+            break
         else:
             # Already mid-flow to something else
             return JSONResponse(


kymckay (changes requested):
I don't think this is right, the actual error returned in this case seems intentional, from my reading of the code this change should result in no visible behaviour difference (the while loop will either break or hit function return - where break is effectively the same as pass from the if statement).

To meet expected user behaviour we probably need to rework the backend logic to do something other than return an error in the case that the user is mid-flow elsewhere.

ajr0d:
> I don't think this is right, the actual error returned in this case seems intentional, from my reading of the code this change should result in no visible behaviour difference (the while loop will either break or hit function return - where break is effectively the same as pass from the if statement).

This is does have a behaviour change, it now lets the user click Google, hit back and use GitLab (and vice versa).  Before, it would keep doing that "400 Bad Request".  I agree with that the actual error is intentional, but I'm not sure in which cases it should happen.

>To meet expected user behaviour we probably need to rework the backend logic to do something other than return an error in the case that the user is mid-flow elsewhere.

I also initially thought this, and that's why I opted for a `while` loop as I thought there could be a way inside the `else` to loop to the top of the `while` statement, but that's obviously not happening so I'm at a bit of a loss of what would be a more approriate workaround while keeping the error.

kymckay:
> This is does have a behaviour change, it now lets the user click Google, hit back and use GitLab (and vice versa). Before, it would keep doing that "400 Bad Request". I agree with that the actual error is intentional, but I'm not sure in which cases it should happen.

I see, I had overlooked that the login dict's method is being assigned and forced to match the method passed as an argument. In that case, in terms of the local logic:

- The `if` branch is unnecessary since it will always be true.
- The `else` branch is unreachable.
- The `while` seems unnecessary because the logic will always break (if `else` was reachable it would break or return) from the loop on the first iteration. It's also unclear why the loop would be necessary if this weren't the case.

In terms of the wider logic my review process is:

1. I need to check if modifying the dict like this is safe as far as the caller is concerned. It smells of an antipattern.
2. The dict is coming from the login state, which looks to be given as an rvalue (for lack of a more Pythonic term) so it's safe to modify.
3. However, by modifying it we're locally overriding the true `"active-login-flow"` retrieved, which seems hacky and like it could have unintended consequences.

It seems to me like the proper way to resolve this is to make sure if already logging in, we just clean that up and then initiate the new flow. That may be what this change is achieving (I'm not super familiar with the rest of the logic that follows in this method), but it's unclear from the code change / justification given here.

danielsilverstone-ct:
We'll close this because it's not ideal as an approach and we should reconsider how to achieve the goals.  Thanks for your attempt though.

PR #259, opened 2022-05-15
Add vcs-browser and contribute URL types

JakobDev:
This PR is the successor of #79

Flathub has now the first Apps which are using the new URL types. A example of a App which uses both new types, so you can test it, is io.qt.Linguist.

grafik


razzeee:
LGTM, but needs formatting

JakobDev:
@razzeee  What command should I run for auto formatting?

razzeee:
`npm run format:fix` https://github.com/flathub/website/blob/main/frontend/package.json#L13

JakobDev:
I have now ran format:fix.

razzeee:
Thank you

PR #260, opened 2022-05-15
Add privacy policy

### LINKED ISSUE #228: Add privacy policy
We need a privacy policy adding to be compliant with various jurisdictions, as we're actually going to be collecting (admittedly very limited) information about people.

razzeee:
Closes #228

Very basic implementation

### DIFF
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -203,5 +203,7 @@
   "cancel": "Cancel",
   "delete": "Delete",
   "go-to-page": "Go to {{page}}",
-  "account-deletion-token-mismatch": "The account deletion request failed to authenticate."
+  "account-deletion-token-mismatch": "The account deletion request failed to authenticate.",
+  "privacy-policy": "Privacy Policy",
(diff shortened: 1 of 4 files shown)


frontend/public/locales/en/common.json
```diff
@@ -203,5 +203,7 @@
   "cancel": "Cancel",
   "delete": "Delete",
   "go-to-page": "Go to {{page}}",
-  "account-deletion-token-mismatch": "The account deletion request failed to authenticate."
-}
+  "account-deletion-token-mismatch": "The account deletion request failed to authenticate.",
+  "privacy-policy": "Privacy Policy",
+  "privacy-policy-description": "By using this website, you agree to the Privacy Policy."
```
JakobDev, line 208: Maybe you should link here to the Privacy Police

razzeee: The text is used for embeds on other websites. So basically the detail text you get on a twitter or facebook embed. It can probably be improved, but I don't think we can have a link.

PR #271, opened 2022-05-16
Fix Apps which has no releases

JakobDev:
Some Apps e.g. org.yuzu_emu.yuzu has no releases, which results in a fronted error. This PR fixes this by not displaying the releases section if there is no release.


razzeee:
Thank you, probably an oversight when I recently changed this.

PR #274, opened 2022-05-16
Fix Gnome return url

nikodunk:
Currently when logging in with Gnome/Gitlab, the user gets redirected back to localhost even when running on beta.flathub.org. This would fix that. 

Probably would make more sense to switch between production and local environments, but couldn't see how to do this easily.


razzeee:
I do not think we wan this change, this needs to be set via env vars on the server and should default to whatever you need for dev, like it is right now.

nikodunk:
Got it! FYI the env var is set wrong on the server right now then. Thanks!

razzeee:
I think it's known, it was merged today and I think our server admin hasn't had time to deal with that yet :)

PR #278, opened 2022-05-18
refactor(frontend): remove excess main containers

kymckay:
Returning a `"main-container"` div within components results in various superfluous containing divs in the final DOM.

If this div is common to every page we should probably insert it at a single point like we do with `main`.

### DIFF
--- a/frontend/pages/userpage.module.scss
+++ b/frontend/pages/userpage.module.scss
@@ -1,7 +1,0 @@
-.userArea {
-  display: flex;
-  flex-direction: column;
-  row-gap: 15px;
-
-  padding: 10px;
-}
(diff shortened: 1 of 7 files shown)


razzeee:
I think we could have a component, that does this embedding, as it should be the default case, but pages like the `about` page will need a way to opt out, as it has the nice graphic at the top, which might be hard to do otherwise.

kymckay:
Hmm, good point about such pages (the app details page is the same). I'm not immediately sure how to intuitively lift the state up such that a child can opt out of the parent's wrapping behaviour, will have to experiment with that.

kymckay:
With the move to tailwind this should probably just become a common component (`ContentRegion` or something) to factor out the frequently used styling classes

kymckay:
As per above, this PR is obsolete

PR #299, opened 2022-05-26
fix(frontend): add next-themes script hash to script-src

sorin-davidoi:
Apply the suggested fix from https://github.com/pacocoursey/next-themes/issues/106#issuecomment-1122012509. The only downside is that if the script changes when the dependency is upgraded, the hash would need to be updated accordingly.

Closes https://github.com/flathub/website/issues/298.


razzeee:
Thank you

Issue #300, opened 2022-05-26
Remove 'unsafe-inline' from style-src


sorin-davidoi:
We use inline styles in about 13 places in the codebase, and this prevents the removal of `unsafe-inline` from the `style-src` Content Security Policy directive. While not as dangerous as the equivalent in `script-src`, we should probably strive to remove `unsafe-inline` by moving these styles to stylesheets.

razzeee:
I'm torn on this, I really like the convenience of having all in one file and don't really like the module approach. 

If I had styling my way, it would probably be via tailwind, which shouldn't trigger the problem.

So I guess I'm saying feel free to fix it, if it really helps.

sorin-davidoi:
`unsafe-inline` still needs to be removed from `style-src`.

PR #313, opened 2022-05-29
Basic tailwind integration

razzeee:
Adding taliwind and using it for all previous inline styles and some other fixes.

Should fix https://github.com/flathub/website/issues/300 and https://github.com/flathub/website/issues/310

Some things:
- Changes paddings and font sizes slightly
- Lists need to marked up (via a class) as lists, if we want bullets. (https://tailwindcss.com/docs/preflight) Not really a problem, but might be something we need to fix for release notes.
- ~~This doesn't touch colors for now~~
- This also slightly changes how the spinner works, hopefully for the better (I still think it's to much choice)


kymckay (approved):
This looks good to me and I'm happy to support the change to tailwind :+1: Do we need buy-in/approval from anyone in particular to go ahead with it (your issue mentioned previous discussion)?

Had been thinking of doing the exact same sizing options for the spinner so thanks for that too :blush:

frontend/tailwind.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  important: true,
+  content: ["./pages/**/*.{js,ts,jsx,tsx}", "./src/**/*.{js,ts,jsx,tsx}"],
+  theme: {
+    extend: {
+      colors: {
+        primary: "var(--bg-color-primary)",
```
razzeee, line 7: not too happy with the names

frontend/tailwind.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  important: true,
```
razzeee, line 2: should only be needed until we phase out all `*.modules.scss` files

razzeee:
- Shadows are also slightly different

I think we want buyin from @barthalion as he basically blocked it last time.

@sorin-davidoi  would this help fix #300 ?

barthalion:
> I think we want buyin from @barthalion as he basically blocked it last time.

I don't remember the details and I don't particularly care as long as it looks good to everyone.

razzeee:
Turns out I misremembered, sorry @barthalion 

https://github.com/flathub/frontend/issues/3#issuecomment-946896611

Turns out I meant @bilelmoussaoui

sorin-davidoi:
> @sorin-davidoi would this help fix https://github.com/flathub/website/issues/300 ?

I think so, as it removes all the inline styles.

PR #321, opened 2022-05-31
Support Projectgroups

JakobDev:
It is possible for Apps to be in Projectgroups. This is used e.g. by KDE instead of the Developer name. This PR adds support for project groups. It's treated as the same as the Developer name.

grafik

grafik


frontend/pages/apps/collection/projectgroup/[projectgroup].tsx
```diff
@@ -0,0 +1,48 @@
+import { GetStaticPaths, GetStaticProps } from "next"
```
Comment (bilelmoussaoui) [inline frontend/pages/apps/collection/projectgroup/[projectgroup].tsx:1]: would rename the url to /group/ instead of /projectgroup/

Comment (razzeee) [reply frontend/pages/apps/collection/projectgroup/[projectgroup].tsx:1]: isn't the important part of info the `project` part? This endpoint basically shows that something belongs to a project. Groups are potentially very different things (and might infact be something we might want to use later for something else)

frontend/package-lock.json
```diff
@@ -0,0 +1,6594 @@
+{
```
sorin-davidoi, line 1: This probably shouldn't be committed, as this project uses Yarn,

frontend/pages/apps/collection/projectgroup/[projectgroup].tsx
Comment (bilelmoussaoui) [reply frontend/pages/apps/collection/projectgroup/[projectgroup].tsx:1]: It all depends on how this is represented in the UI side. But yeah, at least make it `project-group`. `projectgroup` is a bit too much

razzeee:
When I run the backend code, I get a group named `none`, is that a bug?

frontend/public/locales/en/common.json
```diff
@@ -100,6 +100,7 @@
   "rss-description": "Subscribe to RSS feeds from Flathub",
   "rss-applications": "Do you need an RSS application? Find excellent ones on <1>Flathub</1>.",
   "applications-by-developer": "Applications by {{developer}}",
+  "applications-by-projectgroup": "Applications in the {{projectgroup}} group",
```
razzeee, line 103: I kinda want to use the same wording as for developer here, I'm not convinced the `group` part is needed

razzeee: Having played with this a bit, I don't think we can drop the group wording. As flatpak and gnome for e.g. have some entries, that I wouldn't expect and having the wording not use group, makes it sound too official in my mind.

JakobDev:
> When I run the backend code, I get a group named `none`, is that a bug?

No, it's not a Bug. When you look at the 2 Apps, in the none project group ([edu.princeton.physics.WSJTX](https://github.com/flathub/edu.princeton.physics.WSJTX/blob/master/edu.princeton.physics.WSJTX.metainfo.xml) and [io.github.webcamoid.Webcamoid](https://github.com/flathub/io.github.webcamoid.Webcamoid/blob/master/io.github.webcamoid.Webcamoid.metainfo.xml)), both have this line in the metainfo:
```xml
<project_group>none</project_group>
```
I guess this line is there by accident, but it's not a Bug of the Backend.

frontend/pages/apps/collection/projectgroup/[projectgroup].tsx
Comment (JakobDev) [reply frontend/pages/apps/collection/projectgroup/[projectgroup].tsx:1]: I personally find projectgroup without a - in the URL nicer, but I want to hear what others think.

Comment (razzeee) [reply frontend/pages/apps/collection/projectgroup/[projectgroup].tsx:1]: I don't really care, if we're only talking about the url, tbh. It would probably be a bit more consistent, if we do `project-group`, as the spec has a break (underscore) there too https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#tag-ct-project_group

Comment (JakobDev) [reply frontend/pages/apps/collection/projectgroup/[projectgroup].tsx]: I have now renamed the URL

razzeee:
Thank you!

barthalion:
Here's an example why it's a bad idea: https://beta.flathub.org/apps/details/com.clarahobbs.chessclock

Anyone can attach their app to a specific project group, regardless of their actual affiliation.

razzeee:
This might be a bug, the tag isn't in https://gitlab.gnome.org/cghobbs/chess-clock/-/blob/main/data/com.clarahobbs.chessclock.appdata.xml.in

barthalion:
```
~/.l/s/f/a/c/x/s/b/f/s/app-info ❱ zgrep project_group xmls/com.clarahobbs.chessclock.xml.gz
    <project_group>GNOME</project_group>
```

My point stands – anyone can do it. It needs to be tied to first party verification, or we need ID-based heuristic for grouping apps this way.

JakobDev:
> Anyone can attach their app to a specific project group, regardless of their actual affiliation.

That's true, but it is the same for Developer names. I can also put e.g. Microsoft as Developer name for my own App. You can wite into your AppStream what you want. In this case, the App is hosted on Gnome Gitlab, so it has some sort of connection to the Gnome Project.

barthalion:
I said the same thing about developer names, though. Long term, it's not sustainable.

PR #328, opened 2022-06-01
Add random apps collection

JakobDev:
Currently Flathub has 4 Collections on the Homepage: Editors Choice Apps, Editors Choice Games, New & Updated Apps and Popular Apps. The only Collection on which a average App which is not quit popular could land is  New & Updated Apps. But this Collection has the problem, that Apps which got updated often are more often displayed in this Collection on the Homepage. So if you have a cool App that don't get often Updates, the Chance that Users discover the App on Flathub is very low. Even the Explore Page is sorted by installations, which is bad for new and currently not so popular Apps.

To solve this Problem, this PR added a new Collection which contains some random Apps. Each App on Flathub have the exact same Chance to be presented there. This would allow App authors to show their App more people and Users will have the Chance to discover new cool Apps, that the wouldn't find otherwise.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -1,6 +1,6 @@
+import random
 from functools import lru_cache
 from typing import Dict
-import random
 
 import sentry_sdk
 from fastapi import FastAPI, Response


barthalion:
We're discussing what to do with the homepage internally. Adding a random section and keeping all the others is not the solution.

JakobDev:
I hope you find a solution for featuring smaller Apps. I'm still thinking, showing a random App at some point is a goof idea.

bragefuglseth:
Now that we have curation / trending apps, is this still needed?

PR #345, opened 2022-06-06
Separate english locales

razzeee:
This might cause weblate to have problems, as we're removing the source language.


frontend/next-i18next.config.js
```diff
@@ -2,9 +2,10 @@
 
 module.exports = {
   i18n: {
-    defaultLocale: "en",
+    defaultLocale: "en_GB",
```
wjt, line 5: GB as default! :tada:

frontend/pages/statistics.tsx
```diff
@@ -64,7 +65,7 @@ const Statistics = ({ stats }: { stats: Statistics }): JSX.Element => {
                 content: {
                   type: "text",
                   text: stats.downloads?.toLocaleString(
-                    i18n.language.substring(0, 2),
+                    getIntlLocale(i18n.language),
```
wjt, line 68: but I don't think this is going to work. The web platfrom uses the hyphenated form, `en-GB`. But this function appears to return the underscored form.

wjt (commented):
```
new Date().toLocaleDateString("en_GB", {dateStyle: 'long'})
Uncaught RangeError: Incorrect locale information provided
    at Date.toLocaleDateString (<anonymous>)
    at <anonymous>:1:12
(anonymous) @ VM95:1
```

frontend/next-i18next.config.js
```diff
@@ -2,9 +2,10 @@
...
     locales: [
-      "en",
+      "en_GB",
+      "en_US",
```
wjt, line 8: In my experience it is very unusual to have a specific US locale. More typically there is an 'en' one and then some variations like GB, CA, AU, ...

frontend/pages/statistics.tsx
wjt: Oh I see, this is actually returning a `Locale` object.

frontend/next-i18next.config.js
razzeee: Yeah, looking at the formats, I think most of the world will have a much better time with those number/date formats.

frontend/next-i18next.config.js
razzeee: I'm not sure why that would matter, I think the website will route you to the US version if your browser says you would like that.

But yeah, reordering would probably help us avoid the weblate breakage.

frontend/next-i18next.config.js
razzeee: Probably won't last, if we go for the less invasive version for now.

PR #346, opened 2022-06-06
Use long format for date(time) strings

### LINKED ISSUE #329: Make release note date tooltips non-US-specific
On, eg, https://beta.flathub.org/apps/details/org.freedesktop.Bustle, release notes' timestamps are rendered like this, with the exact date in the tooltip:

```
<div title="7/31/2020">almost 2 years ago</div>
```

7/31/2020 is the US-specific M/D/Y format. In this particular case it is unambiguous, but in general this format is ambiguous to a global audience.

My browser sent the following header with my request:

```
accept-language: en-GB,en;q=0.9,fr-FR;q=0.8,fr;q=0.7
```

so the site could adjust to my regional taste for a more logical date format. But failing that / in the absence of l10n in general on the site, a better format might be "31st July 2020".

wjt:
Currently all regional variations of any given language are ignored:
UK and US English are treated as one language.

Sadly the numerical date format is different between the US (m/d/y) and
most of the rest of the anglosphere (d/m/y), and ambiguously so.

Use the "long" format for formatted dates. 'June 5, 2022' may not be the
way yesterday's date would be written in the UK, but unlike '6/5/2022'
it doesn't read as '6th May 2022'.

Fixes #329

### DIFF
--- a/frontend/src/components/application/Releases.tsx
+++ b/frontend/src/components/application/Releases.tsx
@@ -55,6 +55,7 @@ const Releases: FunctionComponent<Props> = ({ latestRelease }) => {
                     latestRelease.timestamp &&
                     new Date(latestRelease.timestamp * 1000).toLocaleDateString(
                       i18n.language.substring(0, 2),
+                      { dateStyle: 'long' },
                     )
                   }
                 >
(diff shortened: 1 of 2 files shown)


razzeee:
Closing this, feel free to reopen if you want to pursue it.

PR #358, opened 2022-06-09
Show Link in Release section

JakobDev:
Programs can provide a Link to more Informations in the Release section of Appstream. This is used e.g. by Inkscape to link to a Wiki page with a more detailed changelog. This PR adds this Link to the Website.

grafik

### DIFF
--- a/frontend/src/components/application/Releases.tsx
+++ b/frontend/src/components/application/Releases.tsx
@@ -32,10 +32,19 @@ const Releases: FunctionComponent<Props> = ({ latestRelease }) => {
     [t],
   )
 
-  const releaseDescription = useMemo(
+  var releaseDescription = useMemo(
     () => (latestRelease.description ? latestRelease.description : noChangelog),
     [latestRelease.description, noChangelog],
   )
+
+  if (latestRelease.url) {
+    releaseDescription +=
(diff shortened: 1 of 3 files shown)


frontend/src/components/application/Releases.tsx
```diff
@@ -32,11 +32,20 @@ const Releases: FunctionComponent<Props> = ({ latestRelease }) => {
...
+  var releaseDescription = useMemo(
     () => (latestRelease.description ? latestRelease.description : noChangelog),
     [latestRelease.description, noChangelog],
   )
 
+  if (latestRelease.url) {
+    releaseDescription +=
+      "<a href='" +
+      latestRelease.url +
+      "' class='list-disc my-4 pl-10'>" +
```
razzeee, line 44: should probably add `target="_blank" rel="noreferrer"` ?

razzeee: Yes, but we probably shouldn't assume current browsers. At least the linting rules we have, don't do that.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

razzeee:
Thanks

PR #367, opened 2022-06-11
Fix docker copy

razzeee:
Deployment seems broken because of this right now. Can you check, if this is what you wanted to do @barthalion


barthalion:
Oops, yeah.

PR #385, opened 2022-06-17
feat(frontend): use modal for deletion confirmation

kymckay:
This PR actually makes the confirmation dialog component use Headless UI's modal component which brings all the benefits of:

- Adding a transition
- Built-in accessibility attributes and navigation
- Inherently using a React portal to render on top of the page and easily centre the dialog

I also moved the account deletion button to the bottom of the user page since it's a one-time action that doesn't contribute to user convenience by being immediately available at the top of the page (and associated with the other actions there).

New deletion button location:
Screenshot from 2022-06-17 18-39-58

Modal dialog in action:
Screenshot from 2022-06-17 18-40-10

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,6 +13,7 @@
     "format:fix": "prettier --write --ignore-path .gitignore ."
   },
   "dependencies": {
+    "@headlessui/react": "^1.6.4",
     "@jonkoops/matomo-tracker-react": "^0.7.0",
     "@stripe/react-stripe-js": "^1.8.1",
     "@stripe/stripe-js": "^1.29.0",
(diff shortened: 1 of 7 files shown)


razzeee:
I think we want the modal to have a lot more spacing, especially outer spacing

kymckay:
More spacing, using dialog title for prompt and replaced the border with some medium shadow:

image

razzeee:
I've pushed another improvement. I'm also kind of wondering, if I would prefer the delete account to be on the right and not on the left

kymckay:
I just did some quick UX research to see if button order is noted as significant anywhere, but doesn't seem to be so no objections from me to switching that around :+1: (everything I did find we seem to be doing well here)

Pretty sure I originally put them in that order just for consistency with some other modals I checked, but I see that Youtube has them reversed for example so that's not a widespread standard.

razzeee:
I think that makes sense

PR #396, opened 2022-06-22
Reimplement the v1 API

barthalion:
The v1 backend fell into obscurity and misses security and bugfixes
updates. Reimplementation on top of the v2 primitives should be more
maintainable and will also yield better search results "for free".

### DIFF
--- a/backend/app/compat.py
+++ b/backend/app/compat.py
@@ -1,0 +1,163 @@
+from datetime import datetime
+from typing import Optional
+
+import requests
+from fastapi import APIRouter, BackgroundTasks, FastAPI
+
+from . import apps, db, search
+
+router = APIRouter(prefix="/compat")
+
+
+def register_to_app(app: FastAPI):
+    app.include_router(router)
+
+
+def get_repo_creation_date(appid: str) -> Optional[str]:
+    key = f"created_at:{appid}"
+    if created_at := db.redis_conn.get(key):
+        return created_at
+    else:
+        try:
+            github_repo = requests.get(f"https://api.github.com/repos/flathub/{appid}")
+            github_repo.raise_for_status()
+            created_at = github_repo.json().get("created_at")
+            db.redis_conn.set(key, created_at)
+            return created_at
+        except requests.exceptions.RequestException:
+            pass
+
+
+def get_short_app(key: str):
+    compat_app = None
+    if app := db.get_json_key(key):
+        appid = key[5:]
+        compat_app = {
+            "flatpakAppId": appid,
+            "name": app.get("name"),
+            "summary": app.get("summary"),
+            "iconDesktopUrl": app.get("icon"),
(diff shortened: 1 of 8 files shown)


backend/app/compat.py
```diff
@@ -0,0 +1,148 @@
...
+def register_to_app(app: FastAPI):
+    app.include_router(router)
+
+
+def get_repo_creation_date(appid: str) -> Optional[str]:
+    if created_at := db.redis_conn.get(f"{appid}:added_at"):
+        return created_at
+    else:
+        try:
+            github_repo = requests.get(f"https://api.github.com/repos/flathub/{appid}")
```
razzeee, line 21: This might lead to us exceeding the api limit, but hopefully not for long until everything is filled.

backend/app/compat.py
```diff
@@ -0,0 +1,148 @@
...
+def get_apps():
+    return list_apps_in_index()
+
+
+@router.get("/apps/category/{category}")
+def get_apps_in_category(category: str):
+    return list_apps_in_index(f"categories:{category}")
+
+
+@router.get("/apps/collection/recently-updated")
```
razzeee, line 77: Other collections will be needed too I guess?

razzeee:
I would like us to have an issue, that tracks the removal of this, somewhen in the near future.

backend/app/compat.py
```diff
@@ -0,0 +1,148 @@
...
+
+def get_release_date(app) -> Optional[str]:
+    if release_ts := app.get("releases")[0].get("timestamp"):
+        release_ts = int(release_ts)
+        return datetime.utcfromtimestamp(release_ts).strftime("%Y-%m-%d")
+
+    return None
+
+
+def get_app_blurb(key: str):
```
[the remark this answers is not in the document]
barthalion: Renamed.

backend/app/compat.py
barthalion: This is the only collection in the v1.

barthalion:
FYI @ahayzen, `inStoreSinceDate` will become the best effort again, based on requests to `/apps/{appid}`. You may need to do slight adjustments to the stats tool.

backend/app/compat.py
razzeee: I would have thought, that there is at least a `new` collection in use https://github.com/flathub/linux-store-backend/blob/da80c7bc96c9d7f6cb30723d7e1cf2e413a0a766/src/main/java/org/flathub/api/service/ApiServiceImpl.java#L21

barthalion: It's used only for the RSS feed: https://flathub.org/api/v1/apps/collection/new/feed

We already implement it in the v2: https://github.com/flathub/website/blob/main/backend/app/main.py#L205=

barthalion:
Migrated `created_at` keys from the old Redis instance so there should be no too many apps missing the field.

PR #405, opened 2022-06-23
fix(frontend): redirect to stored page on login

kymckay:
The effect was running a second time when the stored value was
cleared, which results in default redirection to the user page.

I also updated the page to use the async hook and break the code up for
clarity.

### DIFF
--- a/frontend/pages/login/[service].tsx
+++ b/frontend/pages/login/[service].tsx
@@ -3,35 +3,59 @@ import { useTranslation } from "next-i18next"
 import { serverSideTranslations } from "next-i18next/serverSideTranslations"
 import { NextSeo } from "next-seo"
 import { useRouter } from "next/router"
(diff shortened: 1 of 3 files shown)


frontend/pages/login/[service].tsx
```diff
@@ -3,56 +3,85 @@ import { useTranslation } from "next-i18next"
...
+  }, [t, error])
+
+  // Once router ready, perform login
+  useEffect(() => {
+    if (router.isReady) {
+      execute()
+    }
+  }, [router.isReady, execute])
+
+  // Redirect back to where user was, or userpage for unprompted login
```
Comment (razzeee) [inline frontend/pages/login/[service].tsx:46]: This comment seems to be unrelated to the next line now?

Comment (kymckay) [reply frontend/pages/login/[service].tsx:46]: I figure it pertains to the whole block below, with the state being an implementation detail of that section of code

PR #406, opened 2022-06-23
Limit per-app download stats to the last 180 days

barthalion:
The graph is not very useful for applications available on Flathub for
years, other than showing general trend of Flathub growing.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -225,8 +225,15 @@ def get_stats(response: Response):
 
 
 @app.get("/stats/{appid}", status_code=200)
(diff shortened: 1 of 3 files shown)


razzeee:
I don't have any hard feelings about this, so feel free to merge it. I think it might be nice, to allow to switch between these at some point.

barthalion:
The raw data is still available at https://flathub.org/stats or via API if you pass `all=true`. The main website is a shopfront, detailed data like this belong to developer portal (which we don't have yet).

PR #425, opened 2022-06-30
purchases: Add storefront-info endpoint

jameswestman:
Just a stub endpoint for now. It returns data for flat-manager to include in uploaded commits, including the token-type if purchases/donations are required/accepted.


kymckay:
Do we want to merge this as is or treat as draft until no longer a stub?

PR #435, opened 2022-07-06
feat(frontend): constrain currency inputs

### LINKED ISSUE #305: Cannot cancel a donation of over 1 mil
If someone feels really generous and wants to donate > 1 mil and changes their mind halfway through, a bug occurs:
1. Go to https://beta.flathub.org/donate
2. Enter amount, in my case 1.5 mil
3. Press 'Make Donation'
4. Whoops! Error: Failed to load Stripe.js - happens for every attempt so I guess it's expected
5. Press 'Go to User Wallet'
6. Press 'Cancel Transaction' next to the generous donation

Current behavior:
400 response: `{"status":"error","error":"stripe-payment-intent-build-failed"}`

Expected behavior
We need some input validation on the frontend I think. Do not allow proceeding with bonkers values. When one is even more generous, then other bugs occur:
1. $22000000000000.00 yields 500 response `Internal Server Error` without any message
2. $999999999999999999999999 converts the input to scientific notation 1e+24 and the only way to delete it is by selecting predefined amount
3. If there is a minimal value then it should be validated as well or it is unclear to the user what happens:
https://user-images.githubusercontent.com/105132988/170796432-3b13654f-eed4-458b-871e-aac401e396c6.mov

kymckay:
- Limit maximum input amount to respect the limit Stripe enforces on us.
  The backend will deny any forced requests send above this limit.
- Limit the minimum donation amount to $1

This may be considered to resolve #305 since transactions above the reported value cannot exist any more. Although we may want to only consider that fixed once the cancellation endpoint works regardless of payment intent errors.

### DIFF
--- a/frontend/src/components/application/AppVendingControls/PurchaseControls.tsx
+++ b/frontend/src/components/application/AppVendingControls/PurchaseControls.tsx
@@ -9,6 +9,7 @@ import {
 } from "react"
 import { toast } from "react-toastify"
 import { getAppVendingSetup, initiateAppPayment } from "../../../asyncs/vending"
+import { STRIPE_MAX_PAYMENT } from "../../../env"
 import { useAsync } from "../../../hooks/useAsync"
 import { getIntlLocale } from "../../../localize"
 import { Appstream } from "../../../types/Appstream"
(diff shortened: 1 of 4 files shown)


razzeee (approved):

razzeee:
LGTM code wise

PR #439, opened 2022-07-07
Support GB in calculateHumanReadableSize()

JakobDev:
With RetroDECK we now have a App that has multiple GB (2.5GB install size and 1.1GB download size), so it make sense to show the size in GB. The numbers are displayed with two decimal places: With so big sizes it makes sense to let the User know if the size is 1.01 or 1.99 GB.


razzeee:
Thank you

Issue #256, opened 2022-05-13
Tracking: Donations and payments - WP4


danielsilverstone-ct:
Stuff moved from WP3 around payments and donations

- [ ] Backend support to generate tokens to permit users to be gifted ownership of an app
- [ ] Frontend support for creating and redeeming such tokens
- [ ] The flatpak-authenticator-app workflow needs to be integrated with the workflows
- [ ] Determine tax management approach and implement product upsert and invoice management in stripe backend
- [ ] Frontend support to show ownership information to users
- [ ] Finalised frontend support for defining payment/donation levels

Jamespric:
@danielsilverstone-ct  does this list cover the items mentioned previously in #166 ?
@ramcq Are you happy with this definition for WP4?

PR #441, opened 2022-07-07
Show users their owned applications

kymckay:
- Included owned (purchased) applications in the userinfo endpoint
- Render owned applications on the user page

What I see having setup vending for and then purchased the test app:
Screenshot from 2022-07-07 15-48-31

After deleting my account and recreating it (no longer own the app):
Screenshot from 2022-07-07 15-48-54

This is part of #256

### DIFF
--- a/frontend/src/components/user/UserApps.tsx
+++ b/frontend/src/components/user/UserApps.tsx
@@ -6,13 +6,17 @@ import { useAsync } from "../../hooks/useAsync"
 import ApplicationCollection from "../application/Collection"
 import Spinner from "../Spinner"
 
-const UserApps: FunctionComponent = () => {
+interface Props {
+  variant: "dev" | "owned"
+}
+
+const UserApps: FunctionComponent<Props> = ({ variant }) => {
   const { t } = useTranslation()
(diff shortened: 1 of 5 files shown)


kymckay:
Oh I changed the existing title string for apps the user is a developer of to have clarity between these two sections, but didn't change the associated key because I assume that makes translation easier - do correct me if wrong on that.

frontend/src/components/user/UserApps.tsx
```diff
@@ -25,7 +29,18 @@ const UserApps: FunctionComponent = () => {
     return <Spinner size="m" text={t("loading-user-apps")} />
   }
 
-  return <ApplicationCollection title={t("your-apps")} applications={apps} />
+  const title = t(`${variant === "dev" ? "your" : "owned"}-apps`)
```
razzeee, line 32: I'm unsure if it's worth it to be smart here, being able to find translation strings via the search, might be more valuable in the long run.

kymckay: Good call, I think it will be. Updated.

PR #458, opened 2022-07-13
chore(frontend): update Next.js config entry

kymckay:
The standalone output setting is no longer experimental and is specified like so since version v12.2.0


razzeee:
https://nextjs.org/docs/advanced-features/output-file-tracing#automatically-copying-traced-files

Issue #436, opened 2022-07-06
Handle app vending edge case setup


kymckay:
There is one edge case I think we need to address as a follow up:

- If the developer wants to set the minimum payment to 0 (or any value less than $1.00).
  - Currently the backend will allow this vending setup to be set, but refuse payments of less than $1.00 (because it's too low for the fee calculations).
  - This can be unintuitive as we don't currently explain why this happens to the user on the frontend.

_Originally posted by @kymckay in https://github.com/flathub/website/issues/353#issuecomment-1176055072_

kymckay:
To address this we could:

- [ ] Have the backend refuse minimum payment values less than 1 (unless recommended payment is 0 in which case it doesn't matter since setup is cleared). Just in-case someone manually sends the request.
- [ ] Have the frontend use a toggle to enable/disable all of the vending setup controls (equivalently disabling/enabling vending setup if submit is used).
- [ ] Restrict the vending setup inputs from going below $1.

This should address all the pain points without introducing unintuitive frontend behaviour (e.g. letting you set a value of $0, but not $0.50). It means this restriction is part of the developer story, not the purchaser story (since they will never be presented with an app they can pay $0 for and then be denied).

It does mean we won't support a pay-what-you-want setup where the minimum payment value is 0. If that's a scenario we want to support then there's some additional complexity to be considered.

@danielsilverstone-ct does this align with what we discussed?

kymckay:
There was some confusion around how minimum payment should be treated, clarified as follows:

- When non-zero payment is required for the application.
- When zero, the application can take donations, but no transaction is required for use.

My above PR has been updated to match this understanding and present the frontend payment page differently. The only tweaks needed on the backend:

- [ ] Fix the constraint on the recommended payment column to allow values of $1 (`>=` instead of `>`).
- [ ] Limit both recommended payment and minimum payment to a maximum of 999_999.99 to prevent invalid setups being put into the database (a Stripe enforced limit).

kymckay:
With my open PR for the frontend portion of this, I'm assigning you @danielsilverstone-ct to update/add those backend constraints  (the ones that had to be done manually, see above comment) like discussed at last week's sync.

PR #460, opened 2022-07-13
feat(frontend): add toggle control for app vending setup

kymckay:
Explicitly delineates the "unvended" state via a toggle which disables the other controls, instead of relying on letting the inputs go to $0 to implicitly support unvending an application.

This also means the inputs can be constrained to take minimum values of $1 which the backend should enforce ($1 is the minumum value the backend expects an app to be vended for so that the shares breakdown makes sense with flathub fees in mind).

Edit: I've since updated this to allow the "minimum payment" to be a value of $0, or minimum of $1. ~~This means there's support for a "pay-what-you-want" style of application vending.~~ The UX has suffered slightly (~~feedback now won't disappear until blur and~~ there's no red highlight on the input itself), but that seems less important right now than accommodating the desired behaviour.

Edit2: I've since updated this to present the purchase page as a donation page if minimum payment is $0. The vending setup controls now also include a secondary toggle for requiring payment rather than allowing the input to be $0 or $1+ which was pretty unintuitive and made it unclear how to set up donation only mode.

This is part of the solution to #436. A separate PR is needed to fix the constraint to be `>=` (currently it's `>`) for recommended payment in the backend.


kymckay:
This ended up covering slightly more than intended due to some confusion around how minimum payment should work. It's now functional as follows:

- App vending can be toggled on/off and the form submit to clear/setup vending
- Requiring payment can be toggled on/off as a section of the form, which sets the minimum_payment part of the payload to $0 when off. This allows the input to be constrained to $1 when enabled which is the minimum the backend accepts when it comes to actually making payment.
- If minimum payment is set to $0 then users can donate to an application, an don't need to do so to use it (ownership not required).
- If minimum payment is non-zero than users can pay for an application as usual.

The vending setup page now looks like:
Screenshot from 2022-07-14 16-07-45

When minimum payment is $0 the purchase page becomes:
image

There are some rough edges in the presentation, but they seem low priority right now.

kymckay:
Just some final tidy up of leftover outdated comments. This is now stable for review.

frontend/src/components/application/AppVendingControls/SetupControls.tsx
```diff
@@ -32,6 +33,9 @@ interface Props {
 const SetupControls: FunctionComponent<Props> = ({ app, vendingConfig }) => {
   const { t } = useTranslation()
 
+  const [vendingEnabled, setVendingEnabled] = useState(false)
```
razzeee, line 36: Does the state combination of vending disabled and still requiring a payment make sense? Just a minor nitpick, as I'm curious @danielsilverstone-ct

kymckay: I can answer this :smiley: 

It doesn't, vending disabled (recommended payment of $0) is the master switch which will entirely clear out the setup when submit to the backend. Also as a rule minimum payment cannot be greater than recommended payment which that kind of setup would imply.

razzeee: I would then go for a custom type instead of two states. Might look into this in some days.

PR #464, opened 2022-07-15
Add storybook and some examples

razzeee:
This also adds faker for nicer creation of test data

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ Then you should be able to start the dev server and see changed to the code:
 yarn dev
 ```
 
+### Storybook
+
+You can also start the storybook via:
+
+```sh
+yarn storybook
+```
+
 ## More backend instructions
 
 You can find more backend instructions in the readme in the backend folder.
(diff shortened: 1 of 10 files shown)


kymckay:
Interestingly this works, but I see a lot of error output about broken build due to unresolved models in the terminal.

> ModuleNotFoundError: Module not found: Error: Can't resolve 'fs' in '.../flathub/frontend/node_modules/next-i18next/dist/esm/config'

kymckay (commented):
Interestingly this works, but I see a lot of error output about broken build due to unresolved models in the terminal.

> ModuleNotFoundError: Module not found: Error: Can't resolve 'fs' in '.../flathub/frontend/node_modules/next-i18next/dist/esm/config'

razzeee:
Yeah, I've seen that and investigated it a bit, but didn't feel like it was worth investing more time in.

razzeee:
Tracked the error down and fixed it.

There also was a different one with the a11y plugin, I've removed that for now.

PR #478, opened 2022-07-19
(vending): Add redeemable app tokens API

### LINKED ISSUE #459: Backend support to generate tokens to permit users to be gifted ownership of an app
The goal of this piece is to ensure the backend supports application authors issuing tokens which can be used to redeem ownership of the application by another user.

The detail of the design can be found in this gist: https://gist.github.com/danielsilverstone-ct/ac4791401cbdb4083409baa66a689102

danielsilverstone-ct:
This implements the redeemable token API, it is slightly adjusted from the gist to be better able to return errors in a way consistent with the other APIs.

Assuming this is good, it closes #459


kymckay (approved):
I found no errors and the API looks feature complete!

PR #486, opened 2022-07-24
Berlin style fixes

jimmac:
Couple of papercuts on the stylesheet we found during the Berlin GUADEC


razzeee:
I don't think we want the font part. The next.js docs seem clear how they like fonts to be included https://nextjs.org/docs/basic-features/font-optimization even if we're not using a google font.

jimmac:
I understand it's part of the framework, but it's really stepping all over the founding stones of web design. Moving CSS properties into classes, making fonts only work with javascript.</rant>

I'll separate the font issue.

razzeee:
I'm not sure if that's true, the ssr/static site generation might still
render it fine.

Jakub Steiner ***@***.***> schrieb am Do., 28. Juli 2022,
18:40:

barthalion:
I've looked into font optimization and it just doesn't work at all unless you're using Google Fonts. I'd say let's do whatever works given Next.js stance on it.

razzeee:
yeah, let's just do css fonts for now, as it's unclear, when they will correctly support different font providers. The current implementation probably was due to us hoping it would be soon and then just would work. That's probably not going to happen, as far as I read.

frontend/pages/index.tsx
```diff
@@ -42,14 +42,6 @@ export default function Home({
               </Link>
             </div>
           </div>
-          <div className="hidden md:block">
-            <Image
-              width="400"
-              height="300px"
-              src="/assets/landing.svg"
```
razzeee, line 49: Probably want to delete the svg too then?

razzeee (changes requested):
Just one minor deletion missing

razzeee (approved):

PR #503, opened 2022-07-26
Rework how we display pagination

razzeee:
image

image

@bertob This is a quick iteration on your design. It's much bigger for now, otherwise we would have to do a different version for mobile form factors.

I'm also not sure about the hover being the same effect as the current page.

Closes https://github.com/flathub/website/issues/500

### DIFF
--- a/frontend/src/components/Pagination.stories.tsx
+++ b/frontend/src/components/Pagination.stories.tsx
@@ -1,0 +1,26 @@
+import React from "react"
+import { ComponentMeta } from "@storybook/react"
+import Pagination from "./Pagination"
+
+export default {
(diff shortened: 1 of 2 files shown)


bertob:
> I'm also not sure about the hover being the same effect as the current page.

Hmm no, I was imagining transparent text color for hover, accent color for active.

Looks good to me otherwise :)

razzeee:
Okay, I'm pretty happy now, will check this later on an actual page and probably merge.

razzeee:
They might be a bit to big and they seem a bit too round, but that will probably adjust over time as things in the general design get rounder.

PR #554, opened 2022-08-08
Fetch list data in one go

razzeee:
Hopefully this will help with the size warnings in the build log and maybe with build time


razzeee:
Only removing `draft` to see what CI does

razzeee:
Should be done now, so no more "draft"

Issue #515, opened 2022-07-30
Update ownership token list immediately on creation


kymckay:
A follow up task from #506

Currently if you create new ownership tokens for an app, you have to refresh the page for the list to update with them. This isn't the end of the world, but an end user would expect the new tokens to be immediately shown (intuitive feedback that their action had a successful effect).

To achieve this, the token list itself should be stored as state which can be updated locally to insert the new tokens returned from the successful creation request to the backend.

PR #606, opened 2022-08-24
Tokenlist update

mihaistate03:
Created a solution for issue #515 

This solution fetches the newly created token to the token list after the Submit button is pressed and the page is updated without having to refresh it. If the name of the token is not submitted and the user exits the text field, that name will be deleted.

### DIFF
--- a/frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenList.tsx
+++ b/frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenList.tsx
@@ -1,8 +1,13 @@
 import { Disclosure } from "@headlessui/react"
 import { useTranslation } from "next-i18next"
-import { FunctionComponent, ReactElement, useCallback } from "react"
+import {
+  FunctionComponent,
+  ReactElement,
+  useCallback,
+  useEffect,
+  useState,
+} from "react"
 import { getVendingTokens } from "../../../../asyncs/vending"
-import { useAsync } from "../../../../hooks/useAsync"
 import { Appstream } from "../../../../types/Appstream"
 import Spinner from "../../../Spinner"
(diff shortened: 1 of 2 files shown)


frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenList.tsx
```diff
@@ -18,11 +23,24 @@ interface Props {
...
-  } = useAsync(useCallback(() => getVendingTokens(app.id), [app.id]))
+  const [tokens, setTokens] = useState(null)
+  const [status, setStatus] = useState("pending")
+  const [error, setError] = useState(null)
+
+  const doFetch = useCallback(async () => {
+    try {
+      const fetch = await getVendingTokens(app.id)
+      setTokens(fetch)
+      setStatus("done")
```
razzeee, line 34: The old implementation used `success`, not sure if that matters.

```suggestion
      setStatus("success")
```

frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenList.tsx
```diff
@@ -18,11 +23,24 @@ interface Props {
 const TokenList: FunctionComponent<Props> = ({ app }) => {
   const { t } = useTranslation()
 
-  const {
-    value: tokens,
-    status,
-    error,
-  } = useAsync(useCallback(() => getVendingTokens(app.id), [app.id]))
+  const [tokens, setTokens] = useState(null)
+  const [status, setStatus] = useState("pending")
```
razzeee, line 27:
```suggestion
   const [status, setStatus] = useState<
    "idle" | "pending" | "success" | "error"
  >("idle")
```

razzeee (approved):

PR #607, opened 2022-08-24
Focus the Searchbar by pressing slash (/)

mijorus:


frontend/src/components/layout/Header.tsx
```diff
@@ -135,7 +154,7 @@ const Header = () => {
                             onChange={onChange}
                             value={query}
                             className="block w-full rounded-full bg-bgColorPrimary py-2 pl-10 pr-3 text-sm placeholder-gray-500 focus:border-textPrimary focus:placeholder-gray-400 focus:outline-none focus:ring-1 focus:ring-textPrimary dark:text-textPrimary dark:focus:text-white sm:text-sm"
-                            placeholder={t("search-apps")}
+                            placeholder={'/ '+t("search-apps")}
```
razzeee, line 157: I'm wondering, if we can make this more obvious, via an icon or something.

Is there any prior art, that already has a function like this? Most desktop apps seem to do this on `Control + K` for e.g.

mijorus: The two most common shortcuts are ctrl k and /.

Google, github and MDN use /, the Tailwind documentation uses ctrl + k.

Personally, I am not a big fan of Ctrl k because it interferes with the browser's default shortcuts.

razzeee: Hrm, didn't know google, github and mdn have this. I find googles implementation quite strange. Githubs seems to be the one that communicates it best.

Only thing I'm slightly worried about is, if this causes problems when you enter `/` into other textboxes on the page.

mijorus:
image

there is an icon which suits our needs perfectly, what do you think?

mijorus:
image

smaller looks better IMO

razzeee:
Perfect, was basically hoping for something like that. Could be a few more pixels to the left I think.

mijorus:
image
this looks great to me, balances the smaller size of the icon

mijorus:
chrome shows a litte blue cross icon if the input is type search, I had to hide the icon if the user is filling the input

razzeee:
This seems to play fine with other text inputs.

One more thing, I think we should not show this for mobile devices. As I think this only makes sense on the desktop?

mijorus:
With the current implementation, the icon is hidden for mobile devices (see the md:flex class), but the event listener is enabled; however, mobile devices should't be able to generate keydown events if they are not focusing on a text input

razzeee:
Your right, I just didn't make my window small enough :D

razzeee:
This needs some fixes for RTL languages

mijorus:
> This needs some fixes for RTL languages

We have two options here:

1. use the experimental RTL modifier in Tailwind, which is working flawlessly for me: https://tailwindcss.com/docs/hover-focus-and-other-states#rtl-support, which under the hood reads the `direction` property in the `html` tag
2. increase the right margin for all desktop users, which unfortunately adds some gap when chrome inserts its own "clear all" button
image

Issue #537, opened 2022-08-04
Long app descriptions should be collapsed by default


q30a:
https://beta.flathub.org/apps/details/org.telegram.desktop has an excellent description but it should be collapsed by default imho.

q30a:
Also, half of the description space is empty

PR #609, opened 2022-08-26
Collapsed the long app descriptions

mihaistate03:
Implemented a method to collapse by default all descriptions that are longer
than a specified value. Created a button for extending the description and
one for collapsing it.
Addressed issue #537 

Signed-off-by: Mihai State <mihai.state@codethink.co.uk>

### DIFF
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -249,5 +249,7 @@
   "token-redeem-placeholder": "Enter an ownership token",
   "open-menu": "Open Menu",
   "open-user-menu": "Open User Menu",
-  "your-profile": "Your Profile"
+  "your-profile": "Your Profile",
+  "show-more":"Show More",
+  "show-less":"Show Less"
 }
--- a/frontend/src/components/application/Details.tsx
+++ b/frontend/src/components/application/Details.tsx
@@ -1,6 +1,14 @@
 import { AppHeader } from "./AppHeader"
 import { useMatomo } from "@jonkoops/matomo-tracker-react"
-import { FunctionComponent, useCallback, useMemo, useState } from "react"
+import {
+  FunctionComponent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react"
+import React from "react"
 import { Carousel } from "react-responsive-carousel"
 import { Appstream, pickScreenshot } from "../../types/Appstream"
 import { useTranslation } from "next-i18next"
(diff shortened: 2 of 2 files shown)


razzeee:
I actually think we should not do this

ifeeltiredboss:
Maybe it should be discussed? Is description capped in any way or it can be as long as the developer wants?
I know that Flathub is not obliged to follow others but:
1. Google collapses that https://play.google.com/store/apps/details?id=org.telegram.messenger
2. Apple collapses that https://apps.apple.com/pl/app/telegram/id686449807
3. MS does too https://www.microsoft.com/en-us/p/telegram-messenger/9wzdncrdzhs0

razzeee:
I'm not aware of a limit and there is no mention of that in the spec
https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-description

Btw, how weird MS implementation is, having the description twice on that page.

frontend/src/components/application/Details.tsx
```diff
@@ -214,10 +233,28 @@ const Details: FunctionComponent<Props> = ({
...
+              }}
             />
           </div>
+          {scrollHeight > collapsedHeight && (
+            <button
+              className="w-full rounded-tl-none rounded-tr-none rounded-bl-none rounded-br-none border-t py-3 px-0 font-semibold hover:cursor-pointer hover:bg-colorHighlight dark:border-zinc-600"
+              {...getToggleProps()}
+            >
+              {isExpanded ? (
+                <span>{t("Se less")}</span>
```
razzeee, line 252: Translation keys should be lowercase with no spaces.

This one also has a typo

mihaistate03:
Addressed the suggestion from the review regarding the translation keys in the latest commit.

razzeee:
I think the button needs padding and we can probably get rid of the vertical line.

Here's gnome sofwares take 

image

mihaistate03:
image
Is it ok this button style?

razzeee:
Yeah, feel free to push that, I can iterate on the details somewhen later

backend/pyproject.toml
```diff
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 [tool.poetry.dependencies]
 python = "^3.10"
 fastapi = "^0.80.0"
-uvicorn = "^0.18.2"
+uvicorn = "^0.18.3"
```
razzeee, line 11: Backend change shouldn't be needed, possibly a rebase error?

frontend/public/locales/en/common.json
```diff
@@ -249,5 +249,9 @@
   "token-redeem-placeholder": "Enter an ownership token",
   "open-menu": "Open Menu",
   "open-user-menu": "Open User Menu",
-  "your-profile": "Your Profile"
+  "your-profile": "Your Profile",
+  "see-full-description": "See full description",
```
razzeee, line 253: I don't think this is used

frontend/public/locales/en/common.json
```diff
@@ -249,5 +249,9 @@
...
+  "see-less": "See less",
```
razzeee, line 254: Unused

backend/pyproject.toml
mihaistate03: It is indeed an error, I haven`t make any changes to this file.

razzeee: You probably messed up during rebase, might want to be abit more careful there. I'll go ahead and fix this for now :)

razzeee:
@mihaistate03 I took the liberty to modify this quiet heavily, so you might want to have another look at the source. Anyway, I'll merge it and maybe look into replacing these expandable elements with headlessUI disclosures.

PR #630, opened 2022-09-03
Use new poetry installer

razzeee:
The old installer has been deprecated, so we're using the new recommended way now.
This should hopefully help fixing the issues with newer python dep MRs not building okay.


razzeee:
FYI @barthalion

PR #638, opened 2022-09-04
Limit currency textbox

### LINKED ISSUE #443: Currency input can become scientific notation
If you enter a large value into the currency input component then scientific notation will appear and prevent further input (unless you clear the whole field)

razzeee:
We don't want any higher numbers.

Closes #443


kymckay:
May need to be nine to account for the decimal?

razzeee:
@kymckay good catch, I opted to handle it dynamically now

PR #644, opened 2022-09-05
Add API for EOL rebase

JakobDev:
This PR adds a API for a EOL rebase (Apps changing their ID),

There are 2 new endpoints:
/eol/rebase: A simple dict, which shows which App has a EOL rebase
/eol/oldid: Returns a list with all old ids of a App. At the moment, there is no App that changed the ID more than once, so the implementation is currently simple.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -157,11 +157,8 @@ def get_eol_rebase():
 def get_eol_rebase_appid(
     appid: str,
 ):
-    for new_id, old_id_list in db.get_json_key("eol_rebase").items():
-        if appid in old_id_list:
-            return new_id
-
-    return None
+    if value := db.get_json_key(f"eol_rebase:{appid}"):
+        return value
 
 
 @app.get("/projectgroup")
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -149,15 +149,15 @@ def get_developer(
 
 
 @app.get("/eol/rebase")
-def get_eol_rebase_appid():
-    return db.get_json_key(f"eol_rebase")
+def get_eol_rebase():
+    return db.get_json_key("eol_rebase")
 
 
 @app.get("/eol/rebase/{appid}")
 def get_eol_rebase_appid(
     appid: str,
 ):
-    for new_id, old_id_list in db.get_json_key(f"eol_rebase").items():
+    for new_id, old_id_list in db.get_json_key("eol_rebase").items():
         if appid in old_id_list:
             return new_id
 
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -148,6 +148,16 @@ def get_developer(
     return [app for app in result if app]
 
 
+@app.get("/eol/rebase")
+def get_eol_rebase():
+    return db.get_json_key(f"eol:rebase")
+
+
+@app.get("/eol/oldid")
+def get_eol_rebase():
+    return db.get_json_key(f"eol:oldid")
+
+
 @app.get("/projectgroup")
 def get_project_groups():
     return db.get_project_groups()
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -148,12 +148,18 @@ def get_developer(
     return [app for app in result if app]
 
 
+@app.get("/eol/rebase")
+def get_eol_rebase_appid():
+    return db.get_json_key(f"eol_rebase")
+
+
 @app.get("/eol/rebase/{appid}")
-def get_eol_rebase(
+def get_eol_rebase_appid(
     appid: str,
 ):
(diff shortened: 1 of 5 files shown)


razzeee:
I guess the idea is to add some more urls for the eoled apps and then redirect those? I'm not sure we need both endpoints for that.

JakobDev:
> I guess the idea is to add some more urls for the eoled apps and then redirect those?

Yes. We could also transfer download stats and other things.

> I'm not sure we need both endpoints for that

We need only the second endpoint. 

If a App moves a App ID twice e.g.g swirhing from GitHub to GotLab to it's own domain. In this case, com.github.example.App would point to com.gitlab.exmaple.App as rebase and com.gitlab.exmaple.App would point to com.example.App as rebase.

The 2 API's would return this:

/eol/rebase:
```json
{
    "com.github.example.App": "com.gitlab.example.App",
    "com.gitlab.example.App": "com.example.App"
}
```

/eol/oldid:
```json
{
    "com.example.App": ["com.gitlab.example.App", "com.github.example.App"]
}
```

As there is currently no App which does this, this behaviour is currently not implemented in /eol/oldid, but the API is ready, so it can be implemented when this happens and everything will work without changing other parts of the Website.

/eol/oldid is the only API needed internally, but the data for /eol/rebase needs to be queried anyway, so I thought it would be a nice thing to expose it to someone who might be interested in this.

razzeee:
I actually think we would end up using the first one only, if I remember that part of next.js correctly. It might end up doing multiple redirects in some rare cases, but not sure we should prevent that.

JakobDev:
So we should made this redirects:

apps/details/com.github.example.App -> apps/details/com.gitlab.example.App
apps/details/com.gitlab.example.App -> apps/details/com.example.App

instead of this:

apps/details/com.github.example.App -> apps/details/com.example.App
apps/details/com.gitlab.example.App -> apps/details/com.example.App

Probably that Bug should be fixed in nextJS.

razzeee:
Here's an iteration on your work, feel free to drop my commit again, if you don't like it

JakobDev:
I made 2 changes to your commit: The data are now stored internally in a dict with lists as before. This is a better format for internal use later (I'm planning to also transfer the download stats in a separate PR). I also added a eol/rebase endpoint whihc gives returns the dict as it is. This is not used on the Website (The Website uses your eol/rebase/{appid} endpoint), but we already have the data and this endpoint could be very useful for 3rd party tools (e.g. [flathub-data-viewer](https://github.com/JakobDev/flathub-data-viewer)). I didn't touch your frontend code at all.

barthalion:
Please wait with merging this until I have time to review it.

backend/app/main.py
```diff
@@ -141,6 +141,22 @@ def get_developer(
...
+@app.get("/eol/rebase")
+def get_eol_rebase():
+    return db.get_json_key("eol_rebase")
+
+
+@app.get("/eol/rebase/{appid}")
+def get_eol_rebase_appid(
+    appid: str,
+):
+    for new_id, old_id_list in db.get_json_key("eol_rebase").items():
```
razzeee, line 153: This is exactly what I wanted to get rid of, you've now moved all work from the database to the backend. Which basically leads to loading it completely into memory and scanning it.

JakobDev: Should be fixed with the latest commit

JakobDev:
The code now supports changing the App ID multiple times. Heres a little test for the new while loop:
```python
eol_rebase = {
    "app.one.2": ["app.one.1"],
    "app.one.3": ["app.one.2"],
    "app.one.4": ["app.one.3"],
    "app.one.5": ["app.one.4"],
    "app.two.2": ["app.two.1"],
    "app.two.3": ["app.two.2"]
}

while True:
    found = False
    remove_list = []
    for app_id, old_id_list in eol_rebase.items():
        for new_app_id, check_id_list in eol_rebase.items():
            if app_id in check_id_list:
                eol_rebase[new_app_id] += old_id_list
                remove_list.append(app_id)
                found = True
    for i in remove_list:
        del eol_rebase[i]
    if not found:
        break

print(eol_rebase)
```

There is currently one App on Flathub which benefits from these changes: ind.ie.Gnomit was first rebases as org.small_tech.Gnomit and org.small_tech.Gnomit was later rebased as re.sonny.Commit.

razzeee (approved):
Code looks good, not too happy about the double nesting and the while, but it should work.

barthalion:
Looks fine to me, but seems to need a rebase.

JakobDev:
Sorry that I was so silent. I'm a little bit busy currently. Thanks for merging.

> Code looks good, not too happy about the double nesting and the while, but it should work.

I'm also not happy with it, but I couldn't find a better solution.

PR #652, opened 2022-09-05
Add managed apps

jameswestman:
Adds managed apps to the backend, and applies some changes to the frontend to allow the app management page to accept apps that don't yet have appstream data (as a new managed app wouldn't).

Managed apps must be verified by SSO or website verification at the time they are created. Once you create a managed app, you may no longer delete your account; there is not yet a way to delete a managed app, because that also involves some flat-manager work.

This **should not** be merged until the build token changes are made to flat-manager and existing flat-manager clients are updated, because it allows users to create build tokens, which currently give too much access.


backend/app/managed_apps.py
```diff
@@ -0,0 +1,226 @@
...
+def get_managed_apps(login=Depends(login_state)):
+    """
+    Gets the list of apps managed by this user. Managed apps are those that are uploaded by this user, not community
+    repos from github.com/flathub.
+
+    Note: if an app was recently created, it may not be in the appstream data yet.
+    """
+
+    if not login["state"].logged_in():
+        raise HTTPException(status_code=403, detail="not_logged_in")
```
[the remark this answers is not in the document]
jameswestman: Searching through the repo it looks like we're inconsistent on this.

backend/app/verification.py
```diff
@@ -321,41 +318,60 @@ def _verify_app(appid: str, login, verified: bool):
...
+
+
+def _verify_app(appid: str, login, verified: bool):
+    """Marks a flatpak as verified. This requires the logged in user to have an account with the correct provider and
+    with the correct username, as specified by the app ID.
+
+    Returns: Same as get_verification_status()."""
+
+    if not login["state"].logged_in():
+        raise HTTPException(status_code=403, detail="not_logged_in")
```
ifeeltiredboss, line 334: It is sometimes 403 and sometimes 401, I propose 401 for not_logged_in detail everywhere (for coherenece)
```suggestion
        raise HTTPException(status_code=401, detail="not_logged_in")
```

jameswestman: Sounds good, probably best left to another PR though.

jameswestman:
Closing since this is very out of date, especially with the changes to how verification is done.

PR #683, opened 2022-09-17
Fix slash icon overlap on rtl languages

mijorus:
Addresses https://github.com/flathub/website/pull/607#issuecomment-1249073415 by @razzeee


mijorus:
image

PR #695, opened 2022-09-24
Move updating stats to separate endpoint

barthalion:


razzeee:
Probably should also be documented in the readme, as it mentions the old update endpoint.

PR #705, opened 2022-09-27
Remove editorsChoiceApps, bump APPS_IN_PREVIEW_COUNT to 18

barthalion:


razzeee:
I guess we can't get rid of the endpoint/data.json due to the current page still using that?
Should we mark that endpoint deprecated or as compat somehow?

barthalion:
I will handle `picks` rename in a separate PR. Thanks for fixing formatting!

PR #707, opened 2022-09-28
urlencode searches to fix slash in search causing 404

Dvlv:
Fix for https://github.com/flathub/website/issues/378


razzeee:
A `.` still seems to trigger this problem unfortunatly.

Btw, it would be helpful, if you could work on a different branch in your repo and not `main` :)

PR #713, opened 2022-09-29
Encode urls when searching to prevent 404s

Dvlv:
Fix for https://github.com/flathub/website/issues/378
Old PR was https://github.com/flathub/website/pull/707

Hello, looks like the old PR closed itself when I swapped branch. Sorry for the noise.

I have added manual encoding of a `.`, since encodeURIComponent doesn't do it.

A search like "gitlab.com" works for me, but a search of just `.` or `..` was causing a NormalizeError, so I added them to an exclusion list, as I wouldn't think they would be useful searches anyway.

### DIFF
--- a/frontend/src/components/layout/Header.tsx
+++ b/frontend/src/components/layout/Header.tsx
@@ -75,7 +75,7 @@ const Header = () => {
     e.preventDefault()
     const disallowedQueries = ["", ".", ".."]
     if (!disallowedQueries.includes(query)) {
-      const queryEncoded = encodeURIComponent(query).replace(".", "%2E")
+      const queryEncoded = encodeURIComponent(query).replace(/\./g, "%2E")
       router.push(`/apps/search/${queryEncoded}`)
     }
   }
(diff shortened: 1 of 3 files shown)


frontend/src/components/layout/Header.tsx
```diff
@@ -73,8 +73,10 @@ const Header = () => {
 
   const onSubmit = (e: ChangeEvent<HTMLFormElement>) => {
     e.preventDefault()
-    if (query !== "") {
-      router.push(`/apps/search/${query}`)
+    const disallowedQueries = ["", ".", ".."]
+    if (!disallowedQueries.includes(query)) {
+      const queryEncoded = encodeURIComponent(query).replace(".", "%2E")
```
razzeee, line 78: did you mean to only strip the first `.` or is that by mistake? (the api of that function is weird)

Dvlv: Mistake, sorry. Have pushed a correction.

razzeee:
Thank you!

PR #732, opened 2022-10-12
Move /picks to /compat/picks

barthalion:


backend/app/main.py
```diff
@@ -100,6 +99,11 @@ def update_stats():
     stats.update(apps)
 
 
+@app.post("/update/exceptions")
```
razzeee, line 102: This seems unrelated?

barthalion: Yeah, some leftover from my local repo. I will split it to separate PR in the evening.

razzeee (approved):

PR #741, opened 2022-10-19
Implement app verification

razzeee:


frontend/src/asyncs/app.ts
```diff
@@ -36,6 +41,22 @@ export async function generateUpdateToken() {
   }
 }
 
+/**
+ * Generates an update token for the logged in user.
```
jameswestman, line 45: Forgot to update the comment?

frontend/src/components/application/AppVerificationSetup.tsx
```diff
@@ -0,0 +1,89 @@
...
+                {({ open }) => (
+                  <div>
+                    {methodType.login_name} {methodType.login_provider}
+                  </div>
+                )}
+              </Disclosure>
+            )
+          }
+        })}
+        <Button onClick={() => verifyApp(app.id)}>{t("verify-app")}</Button>
```
jameswestman, line 76: The `verify` endpoint is only for "login_provider" verification. Website verification is automatic.

jameswestman (commented):
We probably want to indicate if the app is currently verified, and if so, how. That's available through the `/verification/{appId}/status` endpoint.

Also, this might be better for a follow-up PR, but it would be nice to add instructions for how to verify an app with each method.

frontend/src/asyncs/app.ts
razzeee: Yes :)

frontend/src/components/application/AppVerificationSetup.tsx
razzeee: Moved inside that code path

razzeee:
@jameswestman should be much better now, but I'm struggling to test this, as I only ever seem to see website verification and that stops working after half an hour or so.

I will probably need to setup a `story` for storybook, to finish this completly

razzeee:
This should be ready now. Texts might need some fine tuning, but nothing, that can't be done in main.

PR #744, opened 2022-10-19
backend: Republish app when vending config changes

jameswestman:
Add a configuration option for the flat-manager build secret and API endpoint. When an app's vending configuration is changed, flat-manager is notified to re-publish the app's commits in both the stable and beta repos.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -46,9 +46,18 @@ class Settings(BaseSettings):
     stripe_secret_key: Optional[str] = None
     stripe_public_key: Optional[str] = None
     stripe_webhook_key: Optional[str] = None
-    flat_manager_secret: str = "c2VjcmV0"
-    update_token_secret: str = "c2VjcmV0"
     env: str = "production"
+
+    # Should match "repo-secret" in flat-manager's config
+    flat_manager_secret: str = "c2VjcmV0"
+    # Should be a unique random secret, not in flat-manager's config
+    update_token_secret: str = "c2VjcmV0"
+    # Should match "secret" in flat-manager's config
+    flat_manager_build_secret: Optional[str] = "c2VjcmV0"
+    # The URL for flat-manager. If present, the backend will use it to queue republish jobs when storefront info
+    # changes. To test in development, set the environment variable FLAT_MANAGER_API=http://host.docker.internal:8080
+    # when running docker-compose.
+    flat_manager_api: Optional[str] = None
 
 
 settings = Settings()
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -1,3 +1,4 @@
+import base64
 import os
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


danielsilverstone-ct (commented):
I think this looks OK, I like the time limit on the JWT, that's a good idea.

I assume there's frontend work to come first though.

jameswestman:
> I assume there's frontend work to come first though.

What frontend work would be needed?

razzeee:
Likely something for a different PR, but it seems, the frontend still works (or tries to work) with the old vending calls, which don't exist anymore.

razzeee:
Nevermind, turns out, that API only exists (gets registered) when you have a stripe api key

jameswestman:
Now that republishing support is merged in flat-manager, this is ready to review.

backend/app/worker.py
```diff
@@ -30,3 +35,44 @@ def update():
...
+        for repo in repos:
+            try:
+                response = session.post(
+                    f"{settings.flat_manager_api}/api/v1/repo/{repo}/republish",
+                    headers={"Authorization": token},
+                    json={"app": appid},
+                )
+
+                if response.status_code != 200:
+                    raise VendingError("republish failed")
```
razzeee, line 75: Might be a cool idea, to write these to postgres at some point and attatch them to the app for the dev to see

razzeee:
Would an admin have to trigger this manually somehow, when they set an app as verified manually?

jameswestman:
Currently yes, but if we end up making an admin/moderator dashboard we could do it automatically.

backend/app/config.py
```diff
@@ -46,9 +46,18 @@ class Settings(BaseSettings):
     stripe_secret_key: Optional[str] = None
     stripe_public_key: Optional[str] = None
     stripe_webhook_key: Optional[str] = None
+    env: str = "production"
+
+    # Should match "repo-secret" in flat-manager's config
     flat_manager_secret: str = "c2VjcmV0"
+    # Should be a unique random secret, not in flat-manager's config
     update_token_secret: str = "c2VjcmV0"
```
razzeee, line 54: It might make sense to change this to illustrate the info from the comment above?

razzeee:
This seems sane to me, only some minor nitpicks, but you might want to wait for somebody who actually likes python :)

backend/app/config.py
jameswestman: Which comment are you referring to?

`update_token_secret` is unrelated to republishing, it's for user downloads. I just added the comment because I was commenting the other items.

razzeee: > # Should be a unique random secret, not in flat-manager's config

says it should be a different secret then the `flat_manager_secret` above, but gives the same example secret :)

jameswestman: Ah ok, that makes sense.

PR #762, opened 2022-10-29
Show verification method on verified badges

razzeee:
Closes https://github.com/flathub/website/issues/750

@jameswestman 
@barthalion 
Please check the wording, not too happy with the login provider one, as it would be clearer, if we had the repo url from my POV.


jameswestman:
It verifies the user, not the repo, so `Verified by @jameswestman on GitHub` would probably work.

PR #803, opened 2022-11-15
verification: Use models for some responses

jameswestman:


backend/app/verification.py
```diff
@@ -104,23 +105,34 @@ def _get_domain_name(appid: str) -> str:
...
+def _check_app_id_error(appid: str) -> ErrorDetail:
+    """
+    Returns an ErrorDetail if the app ID is invalid, or None if it is valid.
+    """
+
+    if not is_valid_app_id(appid):
+        return ErrorDetail.MALFORMED_APP_ID
+
+    if not is_github_app(appid):
+        return ErrorDetail.REPO_DOES_NOT_EXIST
```
razzeee, line 135: Wondering if we might want to make this error github specific, so that it's clear we men the github repo.

PR #823, opened 2022-11-23
Use Orders API to set up tax related content

danielsilverstone-ct:
This work enables tax support in transactions on Flathub.  There are a number of Stripe related setups which have to occur before it can be used, such as opting into the Orders Beta, which is being end-of-lined.  Currently this is not documented in the README because it's not clear if it's possible to opt in any longer (cf. https://stripe.com/docs/orders?locale=en-GB ).  Stripe expect to talk about a new API in the new year.

It's not clear if we're better off integrating this work, or else putting it to one side until the new API is produced.  In theory we could also rework things, gather physical addresses, and restructure into an invoice based charging model in the hope we can integrate that instead of Orders; but that won't be as effective as whatever the new Tax API should be.


danielsilverstone-ct:
Instead we merged #832

PR #833, opened 2022-11-28
Keep download stats on rebase

JakobDev:
If a App changes the ID, it has no all download stats from the previous ID

### DIFF
--- a/backend/app/stats.py
+++ b/backend/app/stats.py
@@ -248,6 +248,52 @@ def update(all_app_ids: list):
                 [i[2] for i in dict.values()]
             )
 
+    # Make sure the Apps has all Keys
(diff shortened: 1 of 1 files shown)


razzeee:
@barthalion do we actually want this?

JakobDev:
I personally think, if a App is rebased, it just means the ID is changed. So all data (except verification) should be transferred to the new ID.

barthalion:
No strong opinion on this from me.

razzeee:
> I personally think, if a App is rebased, it just means the ID is changed. So all data (except verification) should be transferred to the new ID.

Makes me wonder, if vending should be transfered, that might in fact be wanted or not, depending on the case

PR #836, opened 2022-11-29
Add dramatiq-based background task worker

barthalion:


backend/docker-compose.yml
```diff
@@ -53,7 +53,7 @@ services:
       - .:/app:z
 
   db:
-    image: docker.io/library/postgres:12
+    image: docker.io/library/postgres:14
```
razzeee, line 56: Can't we go to 15?

barthalion: We're on 14 now on the production environment.

razzeee: Right, forgot, that this is not the real prod conf

backend/README.md
```diff
@@ -170,3 +170,18 @@ STRIPE_WEBHOOK_KEY=... docker-compose up --build`
 When everything has started, if you visit <http://localhost:8000/docs> you will
 see mention of Stripe if things are working properly.
 
+## Running smoke tests locally
```
razzeee, line 173: This should probably replace the instructions is line 43

PR #883, opened 2022-12-23
Make transaction history nicer

### LINKED ISSUE #642: Make the tables responsive
We either need another Design for this or make it scroll able on mobile.

razzeee:
Closes #642

This doesn't link to the detail page anymore, as we already show all data on here (we request the details, when you click the disclosure)

@kymckay could we drop the detail page completely now? I don't think there is a good usecase for it, unless we need that for some reason.


kymckay:
> could we drop the detail page completely now?

I'd be in favour of it, however the main reason it exists is for the terminal state of the checkout flow. The details page is reachable from is the [checkout component](https://github.com/flathub/website/blob/af69ba8f1ddee277befed62b4299b09de508ca9d/frontend/src/components/payment/checkout/Checkout.tsx) which sends you there on completion or cancellation. Stripe certainly needs somewhere to direct back to when the transaction completes, but there's definitely better options for that.

kymckay (commented):
Haven't actually tested, but code review looks good.

Not marking this review as approve approval since removal of details page will cause 404s from checkout flow, but transaction history changes I approve.

razzeee:
Got rid of the non transaction history parts in this PR, will create an issue for that.

PR #900, opened 2022-12-26
Increase timeout for stats endpoint

razzeee:


backend/app/worker.py
```diff
@@ -10,7 +10,7 @@
 dramatiq.set_broker(broker)
 
 
-@dramatiq.actor
+@dramatiq.actor(time_limit=1000 * 60 * 60)
```
razzeee, line 13: @barthalion FYI

PR #915, opened 2023-01-03
Add sentry tracking for frontend

razzeee:
Tested to work fine, can probably be improved with some more data and release vars etc.

@barthalion


frontend/sentry.client.config.js
```diff
@@ -0,0 +1,19 @@
...
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs"
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN
+
+Sentry.init({
+  dsn:
+    SENTRY_DSN ||
+    "https://b512f563700847e787e978ae1c15133c@o467221.ingest.sentry.io/6610580",
```
barthalion, line 12: That DSN is production one?

frontend/sentry.server.config.js
```diff
@@ -0,0 +1,19 @@
...
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs"
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN
+
+Sentry.init({
+  dsn:
+    SENTRY_DSN ||
+    "https://b512f563700847e787e978ae1c15133c@o467221.ingest.sentry.io/6610580",
```
barthalion, line 12: Same question here

frontend/sentry.client.config.js
```diff
@@ -0,0 +1,19 @@
...
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1.0,
```
barthalion, line 14: I assume we don't want sampling rate to be 1.0 when we deploy this to www.flathub.org? We've got a generous plan from Sentry but I'd prefer not hitting its limits.

frontend/sentry.client.config.js
razzeee: I only see one instance in sentry, not sure if there are separate ones for production?

frontend/sentry.client.config.js
razzeee: I have no idea, will need to read up on it, that's what the setup generated.

frontend/sentry.client.config.js
barthalion: My point is, we probably shouldn't make this URL public, as it's prone to abuse. I will need to land a secret `SENTRY_DSN` in CI/CD.

razzeee: Yes, I also thought about removing this, but I don't think it's private information, as it will end up in the client and you will be able to see, that your browser is sending data there, if you really want to find out.

barthalion:
Fair enough, I guess we can rotate it and make it a secret if it becomes abused.On 3 Jan 2023, at 11:00, Kolja Lampe ***@***.***> wrote:﻿
@razzeee commented on this pull request.



In frontend/sentry.client.config.js:
> +    SENTRY_DSN ||
+    ***@***.***/6610580",

Yes, I also thought about removing this, but I don't think it's private information, as it will end up in the client and you will be able to see, that your browser is sending data there, if you really want to find out.

razzeee:
@barthalion added a `ENABLE_SENTRY` env, that would need to be enabled additionally on prod/staging

PR #942, opened 2023-01-19
compat: Avoid an exception when there are no screenshots

barthalion:


razzeee:
Sentry says, that this is still happening, last 12min ago. So either backend hasn't been deployed since this got merged or it's not fixed.

razzeee:
Sentry seems to suggest, that the code hasn't been updated, according to the stack trace. Release tag is `1fb5bc4f123336c8db2e9e46b0646489199c207d` which seems to be too old.

PR #952, opened 2023-01-24
Use @next/font to serve the Inter font

barthalion:


razzeee:
This seems to not work on prod, but you might need to clear your cache to see it break.

razzeee:
Might also just be me, it starts to break, as soon as I open the firefox dev tools on the beta page. Probably cause I have deactivated the caching in the network tab.

PR #962, opened 2023-01-25
models: Filter dev-flatpaks against apps in appstream

barthalion:


razzeee:
This seems to have broken the smoketests?

PR #963, opened 2023-01-25
verification: Add verification by GitHub organization

### LINKED ISSUE #280: Verification for GitHub organizations
Currently, GitHub verification does not work for apps owned by an organization (e.g. com.github.HypatiaProject.hypatia). In this case, we need to check whether the logged in user is an admin of the organization named in the app ID.

jameswestman:
Users who are admins of the named GitHub organization may verify the app.

Fixes #280.


backend/app/verification.py
```diff
@@ -295,17 +312,102 @@ def get_available_methods(appid: str, login=Depends(login_state)):
 
     if provider := _get_provider_username(appid):
         provider_name, username = provider
+        organization = False
+
+        if provider_name == LoginProvider.GITHUB:
+            try:
+                user = github.Github().get_user(username)
```
razzeee, line 319: Shouldn't we be able to use `get_user()` instead or are we not logged in as that user?
Or is this an unauthorized call? If yes it might make sense to authorize due to api limits

backend/app/verification.py
```diff
@@ -295,17 +312,102 @@ def get_available_methods(appid: str, login=Depends(login_state)):
...
+    # For GitHub, we allow verifying by either user or organization. If it's by organization, the user must be an
+    # admin of that organization.
+
+    account = models.GithubAccount.by_user(sqldb, account)
+
+    if account is None:
+        raise HTTPException(status_code=401, detail=ErrorDetail.NOT_LOGGED_IN)
+
+    try:
+        gh = github.Github()
```
razzeee, line 355: Should we do this authorized too?

backend/app/verification.py
jameswestman: `username` is from the app ID. We won't be logged in under that name if it belongs to an organization. We could still authorize the call with the current user's token, though.

jameswestman: I changed some of this code around when I implemented the frontend.

frontend/public/locales/en/common.json
```diff
@@ -254,7 +254,14 @@
...
   "login-provider-verification-switch-account": "You are currently logged in to <1>{{login_provider}}</1> as <3>@{{current_name}}</3>. Log in as <5>@{{required_name}}</5> to verify <7>{{appId}}</7> by login provider, or choose another method if one is available.",
+  "login-provider-verification-not-org-admin": "You are currently logged in to <1>{{login_provider}}</1> as <3>@{{current_name}}</3>. This account is not an admin of the <5>@{{organization}}</5> organization. To verify this app, you must request admin privileges in the organization or sign into an account that already has them.",
+  "login-provider-verification-not-org-member": "You are currently logged in to <1>{{login_provider}}</1> as <3>@{{current_name}}</3>. This account is not a member of the <5>@{{organization}}</5> organization. To verify this app, you must request admin privileges in the organization or sign into an account that already has them.",
+  "login-provider-verification-failed": "The verification requirements were still not met.",
```
razzeee, line 262: This sounds a bit passive aggressive, but I'm not a native english speaker

razzeee:
Code and storybook look good, could not test it locally due to not having an app that qualifies.

frontend/public/locales/en/common.json
jameswestman: Hmm, yeah. The idea was to give feedback when you click Try Again, since the status message above the button hasn't changed. But it can probably just say "The verification requirements were not met."

PR #1008, opened 2023-02-08
Drop /details from applications URLs

barthalion:


razzeee:
Can we redirect in code?

Also the sitemap generator needs an update

PR #1021, opened 2023-02-09
WIP: tweak UI of verification status

barthalion:


barthalion:
Current screenshots:

image

image

barthalion:
The downside of the above idea is that the location of the checkmark implies that the application name has been verified, which is not true. Some people also feel strongly that we shouldn't display the ID directly on the page. Another idea from yesterday:

image
image

This moves the checkmark to the "by John Doe" line. That's in line with what Snapcraft and VSCode Marketplace do. The problem is that it may cause confusion for apps of the same developer where only one is verified. A potential example would be Steam (maintained by nanonyme) and Steam Link (official by Valve).

barthalion:
We could change the tooltip content to explicitly mention the app ID, e.g. "The ownership of the ${ID} application ID has been manually verified by Flathub staff" and "The ownership of the ${ID} application ID has been verified via GitHub by ${username}".

barthalion:
James suggested something similar to the second variant yesterday.

image

This is modelled after website verification feature in Mastodon. I'm just not convinced it conveys the same message.

barthalion:
Felix suggested this:

image

ramcq:
As something that's unambiguously accurate and very non-confusingly displays what we have verified, without the RDNN coming through, I really like James' idea of showing the checkmark next to the actual thing we have verified - ie that we have checked the domain or the GitHub/GitLab account or organisation.

ramcq:
I guess if we have done some additional manual verification or something it might need to say Verified by Staff ☑️ in that fallback case.

razzeee:
As the other PR targets this branch, we should probably keep this open

barthalion:
Oh, I didn't notice that. Thanks

PR #1032, opened 2023-02-10
verification: Tweak verification status UI

jameswestman:
Building on #1021, fix the alignment of the verified badge, fix the missing variable in the tooltip translation, and make the text a hyperlink.


barthalion:
Can you change the tooltip to use human-readable name of the login provider?

image

PR #1039, opened 2023-02-11
Add categories to meilisearch and get category listings from there

razzeee:


backend/app/search.py
```diff
@@ -27,6 +29,24 @@ def delete_apps(app_id_list):
...
+    category_list = [
+        f"categories = {category.value}" for category in selected_categories
+    ]
+
+    return client.index("apps").search(
+        "",
+        {
+            "filter": [category_list],
+            "sort": ["installs_last_month:desc"],
+            "hitsPerPage": hits_per_page or 250,
```
razzeee, line 44: This might be problematic, if somebody ever tries to consume this without a limit, but all our pages have paging right now.

razzeee: Hrm, but we're doing paging client side right now, so this breaks it. Needs more work.

PR #1043, opened 2023-02-12
Keep generating old urls for feeds

razzeee:
As the current live page uses the same backend, we need to switch this over later.


razzeee:
We should prepare a revert MR for this

razzeee:
Closes https://github.com/flathub/website/issues/1031

PR #1045, opened 2023-02-13
New color handling

razzeee:
We'll this was some work.

But after all, the only thing that seems to have slightly changed is one of the black tones in the light variant.

I've kept the linter formatting in it's own commit, as it might be helpful to be able to figure stuff out later, to have the previous commit.

I've reviewed this quiet a bit from a user POV, but I'll like to do a normal code review in the next days.

Basically what this means, is that we can use colors more freely and can use the transparencies of them via tailwind.
See https://tailwindcss.com/docs/customizing-colors#using-css-variables as a reference.

### DIFF
--- a/frontend/pages/about.tsx
+++ b/frontend/pages/about.tsx
@@ -12,7 +12,7 @@ const About = () => {
       <div className="flex flex-col">
         {/* <!-- header --> */}
         <header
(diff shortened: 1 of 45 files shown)


razzeee:
Reviewed this again, should be good to go.

PR #1053, opened 2023-02-13
Restore scroll position on back navigation

### LINKED ISSUE #1052: Remember scroll position when going back
When opening an app from the home page and then going back the scroll position is currently lost (Firefox on Android and on desktop). This is especially annoying on mobile where there's more scrolling.

razzeee:
Closes #1052


razzeee:
Taken from here https://github.com/vercel/next.js/issues/20951#issuecomment-987252541

The topic also mentions some issues with it, I'll try to read through these carefully later and decide if we can proceed like this.

PR #1089, opened 2023-02-17
Add button/endpoint to refresh dev-flatpaks

jameswestman:
If someone is given access to a flathub/* repo on GitHub, they can click Refresh instead of logging out and back in. We can also use the function in other places if we want to make sure the repo list for a user is up to date.

### DIFF
--- a/frontend/src/components/application/Collection.tsx
+++ b/frontend/src/components/application/Collection.tsx
@@ -33,17 +33,6 @@ const ApplicationCollection: FunctionComponent<Props> = ({
 }) => {
   const { t } = useTranslation()
   const router = useRouter()
-  if (!page) {
-    page = parseInt((router.query.page ?? "1") as string)
-  }
-  if (!totalPages) {
-    totalPages = Math.ceil(applications.length / perPage)
-  }
-  const pages = Array.from({ length: totalPages }, (_, i) => i + 1)
-
-  const pagedApplications = totalHits
-    ? applications
-    : applications.slice((page - 1) * perPage, page * perPage)
 
   const refresh = onRefresh ? (
     <Button onClick={onRefresh}>{t("refresh")}</Button>
@@ -68,6 +57,18 @@ const ApplicationCollection: FunctionComponent<Props> = ({
       </div>
     )
   }
+
+  if (!page) {
+    page = parseInt((router.query.page ?? "1") as string)
+  }
+  if (!totalPages) {
+    totalPages = Math.ceil(applications.length / perPage)
+  }
+  const pages = Array.from({ length: totalPages }, (_, i) => i + 1)
+
+  const pagedApplications = totalHits
+    ? applications
+    : applications.slice((page - 1) * perPage, page * perPage)
 
   return (
     <div className="flex">
--- a/frontend/src/components/application/Collection.tsx
+++ b/frontend/src/components/application/Collection.tsx
@@ -1,14 +1,14 @@
-import { FunctionComponent } from "react"
 import { useRouter } from "next/router"
+import { FunctionComponent } from "react"
 
 import { Appstream, AppstreamListItem } from "../../types/Appstream"
 
-import ApplicationCard from "./ApplicationCard"
-import Pagination from "../Pagination"
 import { useTranslation } from "next-i18next"
-import ButtonLink from "../ButtonLink"
 import { UserState } from "src/types/Login"
 import Button from "../Button"
+import ButtonLink from "../ButtonLink"
+import Pagination from "../Pagination"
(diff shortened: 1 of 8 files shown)


frontend/src/components/user/UserApps.tsx
```diff
@@ -13,13 +15,20 @@ interface Props {
...
   const getApps = useCallback(
     () => getAppsInfo(user.info[`${variant}-flatpaks`]),
     [user.info, variant],
   )
   const { execute, status, value: apps } = useAsync(getApps, false)
 
+  const doRefreshDev = useCallback(async () => {
+    await refreshDevFlatpaks()
+    await getUserData(userDispatch)
+  }, [execute, userDispatch])
```
razzeee, line 29: Does this really need to depend on `execute`?

frontend/src/components/user/UserApps.tsx
```diff
@@ -1,9 +1,11 @@
 import { useTranslation } from "next-i18next"
 import { FunctionComponent, useCallback, useEffect } from "react"
-import { getAppsInfo } from "../../asyncs/app"
-import { useUserContext } from "../../context/user-info"
+import { getUserData } from "src/asyncs/login"
+import { getAppsInfo, refreshDevFlatpaks } from "../../asyncs/app"
+import { useUserContext, useUserDispatch } from "../../context/user-info"
 import { useAsync } from "../../hooks/useAsync"
 import ApplicationCollection from "../application/Collection"
+import Button from "../Button"
```
razzeee, line 8: I don't think you ended up using this

frontend/src/components/application/Collection.tsx
```diff
@@ -42,11 +45,38 @@ const ApplicationCollection: FunctionComponent<Props> = ({
...
+  ) : null
+
+  const header = (
+    <span className="flex items-center justify-between">
+      <h2>{title}</h2>
+      {refresh}
+    </span>
+  )
+
+  if (applications.length === 0) {
```
razzeee, line 61: This won't work, as we return early above, if we have no application data.

frontend/src/components/application/Collection.tsx
```diff
@@ -42,11 +45,38 @@ const ApplicationCollection: FunctionComponent<Props> = ({
     ? applications
     : applications.slice((page - 1) * perPage, page * perPage)
 
+  const refresh = onRefresh ? (
+    <Button onClick={onRefresh} className="float-right">
+      {t("Refresh")}
```
razzeee, line 50:
```suggestion
      {t("refresh")}
```

frontend/src/components/application/Collection.tsx
jameswestman: Where is the early return? I moved the "No applications to show" message to ApplicationCollection so I wouldn't have to duplicate the refresh button.

frontend/src/components/application/Collection.tsx
```diff
@@ -42,11 +45,38 @@ const ApplicationCollection: FunctionComponent<Props> = ({
     ? applications
     : applications.slice((page - 1) * perPage, page * perPage)
 
+  const refresh = onRefresh ? (
+    <Button onClick={onRefresh} className="float-right">
```
razzeee, line 49: I don't think `float-right` is needed, cause you move it with `justify-between` below

frontend/src/components/application/Collection.tsx
razzeee: Line 24 in my the edition I've checked out from your branch

frontend/src/components/user/UserApps.tsx
```diff
@@ -13,13 +14,20 @@ interface Props {
...
 
   const getApps = useCallback(
     () => getAppsInfo(user.info[`${variant}-flatpaks`]),
     [user.info, variant],
   )
   const { execute, status, value: apps } = useAsync(getApps, false)
 
+  const doRefreshDev = useCallback(async () => {
+    await refreshDevFlatpaks()
+    await getUserData(userDispatch)
```
razzeee, line 27: I don't really like, that we basically login the user again due to this. I suspect, this is why it seems like the whole page reloads as presumably we're logged out/loading login for a few sec.

It would be much nicer to just replace the `apps` list, but that might be too much to ask? :)

razzeee:
Looks great now, code wise, only the empty state fix is missing and I haven't tested the latest iteration

frontend/src/components/application/Collection.tsx
razzeee: I think I confused some files? Can't find the place anymore.

PR #1090, opened 2023-02-17
verification: Add general instructions

### LINKED ISSUE #1073: Add info box to developer settings about trademark infringements/misleading app names

jameswestman:
Add general instructions describing what verification does, and add an info box explaining that misleading badges may be removed.

Fixes #1073.


frontend/public/locales/en/common.json
```diff
@@ -246,6 +246,8 @@
   "more-type": "More {{type}}",
   "verified": "Verified",
   "app-is-verified": "This app is verified",
+  "verification-instructions": "Verify your app ID to indicate that your Flathub package is approved by the upstream developer. You'll need to confirm your identity, for example with a website or an account on a source code hosting site. The verified badge will be shown alongside the app name, publisher name, and the identity you use for verification.",
+  "verification-warning": "We may remove the verification badge from an app if it could mislead people about the app's official status.",
```
razzeee, line 250: We might want to keep this more general, just to be able to act on certain things? 

@ramcq for wording guidance

frontend/src/components/application/AppVerificationControls/SetupControls.tsx
```diff
@@ -144,6 +145,9 @@ const AppVerificationSetup: FunctionComponent<Props> = ({ app }) => {
   } else {
     content = (
       <div className="space-y-3">
+        <p>{t("verification-instructions")}</p>
```
razzeee, line 148: Usually you don't want these to go too much into the width, as it get's hard to read.

I would recommend `xl:max-w-[75%]` at least. We do already use that in a few other places (or similar things)

For e.g. I like to use `prose dark:prose-invert` for text blocks, but that sometimes feels a bit to aggressive as it cuts down on the width hard. (probably for good reason)
https://tailwindcss.com/docs/typography-plugin#overriding-max-width

frontend/public/locales/en/common.json
```diff
@@ -246,6 +246,8 @@
   "more-type": "More {{type}}",
   "verified": "Verified",
   "app-is-verified": "This app is verified",
+  "verification-instructions": "Verify your app ID to indicate that your Flathub package is approved by the upstream developer. You'll need to confirm your identity, for example with a website or an account on a source code hosting site. The verified badge will be shown alongside the app name, publisher name, and the identity you use for verification.",
```
ramcq, line 249: Verify your app ID to indicate that your Flathub upload is approved by the app developer. You'll need to prove that you have control of the app via a website on the app's domain or by checking access to the app's account on a source code hosting site. The verified badge will be shown alongside the app name, publisher name, and the way that the app ID was verified.

frontend/public/locales/en/common.json
ramcq: The Flathub team may remove the verification badge from any app at our sole discretion, for example if it could mislead people about the app's official status or cause confusion regarding a trademark.

frontend/src/components/application/AppVerificationControls/SetupControls.tsx
jameswestman: How about

image

PR #1100, opened 2023-02-20
Remove category index, as it's unused

razzeee:
We moved the implementation over to meilisearch, so this bookkeeping is not needed anymore


razzeee:
You might want to delete `categories:index` after this

razzeee:
@barthalion

PR #1109, opened 2023-02-22
Dark mode-friendly favicon

akatiggerx04:
I updated the favicon and added support for dark mode. The new icon will appear white in dark mode and black in light mode, making it more visible in the tab bar. The implementation of dark mode support is done using a simple CSS media query  (a nice little touch). Additionally, to optimize the SVG, I used Inkscape's Optimized SVG format instead of the previous plain Inkscape SVG format.


razzeee:
This already works before this PR for me. At least in Firefox and I would think I would have taken note of Gnome Web not doing it too.

Did you refresh the favicon correctly? F5 + Ctrl

Which Browser did you test?

akatiggerx04:
Yes i did refresh correctly, i tested it in both Brave & Librewolf.

razzeee:
Seems like a chromium thing, let me run some more tests.

razzeee:
Firefox and librewolf seem happy. Theme selection in chromium and brave is too confusing for me, but I assume it should work.

Gnome Web however, seems unhappy, it doesn't seem to show the favicon at all. Not sure if due to format or size :/

akatiggerx04:
Indeed! It seems that Gnome Web or probably webkit is having some issues displaying SVG favicons. Even the GitHub and current beta Flathub favicon are not showing up. One possible solution is to include a PNG fallback?

razzeee:
I could only get it to work with ico, but I'm not sure how to do that without affecting the svg icon path

akatiggerx04:
I couldn't get any site with an svg favicon to work in Gnome web; it's probably a bug on their side. Adding a png fallback like this seems to fix the issue and display the fallback image:

```html
<link rel="icon" href="/favicon.svg" sizes="any" type="image/svg+xml">
<link rel="icon" href="/favicon.png" type="image/png">
```
While the svg favicon continues to work in Chromium and Firefox, Gnome web displays the png image.

razzeee:
But the png one would need to have some kind of background to work with both black and white background. At least for legacy browsers.

image


@jimmac

akatiggerx04:
How about using a colored favicon for the fallback image?

Screenshot from 2023-02-23 01-03-44

jimmac:
As the PNG is only a fallback for older webkit, going with a 50% grey may be good enough contrast to remain legible. I would really prefer to keep the logomark neutral.

razzeee:
Gnome Web still seems to not respect that fallback icon as far as I can see. But I don't think it's worth, to investigate this further. 

Thanks for sending the PR.

PR #1111, opened 2023-02-22
Update flathub-logo-mini.svg

DartDeaDia:
Thicker "+", which looks better from a distance.


razzeee:
@jimmac

jimmac:
The + is purposely thinner weight because it's unlike the other symbols resembling app icons. It's meaning can be attributed to more apps coming soon, or inviting 3rd party app developers to list theirs.

PR #1112, opened 2023-02-22
Update flathub-logo-toolbar.svg

DartDeaDia:
Flathub logo with more bold font, that's more similar to the Flatpak logo. Also with thicker "+" as in #1111


razzeee:
@jimmac

jimmac:
The logotype is balanced to the logomark. There is no benefit in making the logo the same or similar to flatpak. See https://github.com/flathub/website/issues/301

PR #1123, opened 2023-02-24
Cleanup search index

razzeee:


backend/app/worker.py
```diff
@@ -20,6 +20,7 @@ def update_stats():
 
 @dramatiq.actor
 def update():
+    search.delete_all_apps()
```
razzeee, line 23: @barthalion this is only temporary, when the search index was reset I will send another commit removing this part again

Issue #1153, opened 2023-02-28
Add verified timestamp to appstream via hook


razzeee:
To make verified appstream data usable for desktop frontends and for us when listing apps, we would need a timestamp of verification. That would allow us to sort by, like we currently do right now on the landing page.

Having these data in appstream would allow us to regenerate more from it and use it with our search engine. We could set up facetted searches or use the search index as backbone for the verified listings.

razzeee:
This has been added

PR #1155, opened 2023-02-28
storefront-info: Add verified timestamp

jameswestman:


backend/tests/main.py
```diff
@@ -617,7 +617,10 @@ def __call__(self, _appid: str, _token: str):
         "website": "gnome.org",
     }
     assert response.status_code == 200
-    assert response.json() == expected
+    json = response.json()
+    assert json["verified"] == True
```
razzeee, line 621: `expected` is unused now?

razzeee:
Seems sane to me.

Should be followed by https://github.com/flathub/flat-manager-hooks/pull/6/files and would close #1153 then

PR #1172, opened 2023-03-02
Add verification data to meilisearch

razzeee:
This will start indexing the verified data we have right now into meilisearch and show a basic checkmark on listings, that come from meili (categories for now)


razzeee:
I have no clue, how this would break stats Oo
Help.

razzeee:
Well, somehow the main branch broke I guess? I can't get it to pass locally at least too. Guess it broke somewhere in the last 18 hours when the last main branch change build and now?

PR #1176, opened 2023-03-02
Fix stats tests

razzeee:
We don't process today, so create data for three days before today

@barthalion I'm very confused why this is or if this will flop back to not working in a few days


razzeee:
Somehow this also doesn't trigger CI, these are strange days

PR #1179, opened 2023-03-04
Implement listings being consuming data from meilisearch

razzeee:
This should mostly work now.

Needs to be double checked, if it matches the current orders, but that can only be done when merging and checking against the current pages.

This also blindly implements the new ordering by verification timestamp, but as that does not show up yet, it's untested and might not work.

So ordering of verified items will be wrong probably.


razzeee:
Right before merge:

image

image

image

image

image

image

PR #1187, opened 2023-03-05
Setup ruff, configure it and fix some problems

razzeee:
Replace pylint, isort and flake. 

Ruff should cover all the above and more. It's used by fastapi and other big projects, so it seems safe enough.

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -33,9 +33,7 @@ dramatiq = {extras = ["redis", "watch"], version = "^1.14.1"}
 [tool.poetry.dev-dependencies]
 black = "^23.1"
 pytest = "^7.2.1"
-isort = "^5.12.0"
-pyflakes = "^3.0.1"
-pylint = "^2.16.2"
+ruff = "^0.0.254"
 
 [tool.poetry.group.dev.dependencies]
 httpx = "^0.23.3"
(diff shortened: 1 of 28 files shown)


backend/pyproject.toml
```diff
@@ -44,5 +42,7 @@ httpx = "^0.23.3"
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
-[tool.isort]
-profile = "black"
+[tool.ruff]
+ignore = ["E501"]
+exclude = ["alembic/versions/*"]
+extend-select = ["I", "SIM", "W", "UP", "TCH"]
```
razzeee, line 48: I = isort
F is on by default and stands for flake see https://beta.ruff.rs/docs/rules/#pyflakes-f

razzeee:
Most of the code changes have been automatic fixes, but I have reviewed all and they at least seem to make sense

backend/pyproject.toml
```diff
@@ -44,5 +42,7 @@ httpx = "^0.23.3"
 requires = ["poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
 
-[tool.isort]
-profile = "black"
+[tool.ruff]
+ignore = ["E501"]
```
razzeee, line 46: We're ignoring the line too long error. We did not enforce this before and it would cause some problems.

barthalion:
the lock file needs regeneration

razzeee:
Rebased/regenerated

PR #1234, opened 2023-03-11
verification: Implement GitLab org verification

jameswestman:
You can now verify an app ID by logging into a GitLab (or GNOME GitLab) account that has owner or maintainer access to the organization named in the app ID.

~~This PR does not add support for `org.gnome.*` app IDs yet.~~


backend/app/verification.py
```diff
@@ -447,13 +448,54 @@ def _verify_by_github(username: str, account) -> bool:
...
+            headers={"Authorization": "Bearer " + access_token},
+        )
+
+        if not r.ok:
+            raise HTTPException(status_code=500, detail=ErrorDetail.PROVIDER_ERROR)
+
+        userinfo = r.json()
+
+        # Must be owner or maintainer of the group
+        if groups := userinfo.get("https://gitlab.org/claims/groups/owner"):
```
razzeee, line 488: Not sure I get this, but will this work for the gnome/kde/freedesktop instances?

jameswestman:
I've added the code for org.gnome.* app IDs, but the app needs openid access to test it.

razzeee:
This seems sane enough to merge and hopefully barth has enough rights on that gitlab instance to verify if it works then.

razzeee (approved):

Issue #906, opened 2023-01-02
[Feature Request] Separate "media consumption" from "media production" apps in the categories


hyuri:
Hi!

Any way to separate the "media consumption" apps from the "media production" ones? (Audio & Video)

I'm trying to look through the media production apps available (and I usually check again from time to time) and it's really hard to distinguish because there are so many media players and other media consumption apps mixed together with the production apps into a long multi-pages list.

razzeee:
Nothing we have right now, but there are additional categories in the spec. Not sure they would be helpful for your usecase https://specifications.freedesktop.org/menu-spec/latest/apas02.html

PR #1240, opened 2023-03-12
Add subcategories for games and audio/video

### LINKED ISSUE #484: Break out the games category into more defined sections
Right now "Games" is 13 pages of launchers, games, utilities, and all sorts of stuff. I think it might be useful to break this out for a few reasons:

From observing Youtubers covering linux gaming and steamdeck stuff there's a lot of "ok search for this, then install, then search for this, then install, then search for this, then install." I'd like to propose splitting "Games" into something more granular. Something like this, I'm sure everyone will have an opinion on naming so I'll just go with large buckets:

"Games" - Actual games
"Emulators" - and their associated utilities
"Gaming Tools" - everything gaming related that is not a game or emulator, launchers, etc.

and then lastly, I think having a "Core Gaming" thing would be useful, basically so when someone is getting into linux gaming people can just say "you'll probably need all of these, just install everything in this category" So a small curated list, here's a first stab: 

```
com.valvesoftware.Steam
com.valvesoftware.Steam.Utility.gamescope
org.freedesktop.Platform.VulkanLayer.MangoHud
com.heroicgameslauncher.hgl
org.openrgb.OpenRGB
com.usebottles.bottles
com.github.Matoking.protontricks
```

I am unsure if categories can be nested or if they'd all be top level, so filing this for discussion. When we have a rough consensus I'd be happy to whip up a PR if someone could outline how new categories are made.

razzeee:
Closes #484 #906


razzeee:
We might want to hide some of the weirder categories such as `telephony tools`?

razzeee:
Rebased and changed to game & audio/video instead of game & utilities

Might allow closing https://github.com/flathub/website/issues/906

razzeee:
This is missing the needed sitemap changes

PR #1241, opened 2023-03-12
Add a user collection

### LINKED ISSUE #793: Add a bookmarks system for app
I find that there is a account system on beta.flathub.org, so I think it will be nice to have a bookmarks system for apps to make find apps again easily

razzeee:
Closes #793

This allows you to add apps to your personal collection or remove them.

We should also list the collection in the users private profile.

I'm not too sure about the naming `collection`, but I don't like `bookmark` either.

The frontend is pretty rought/unfinished. The backend works, as far as I can tell, but not a python crack.


JakobDev:
If we are adding this, I think it would be a good idea, to allow multiple lists which custom Names, so a User could maybe have a Development Tools and a Games list.

razzeee:
I also thought about public lists etc, but that's not the clear design I would want.

I think a collection is different from a custom list. Mostly taken Inspiration from pages like Trakt.

Bekon241:
> If we are adding this, I think it would be a good idea, to allow multiple lists which custom Names, so a User could maybe have a Development Tools and a Games list.

Hi, or just one list but you can add text box with your own discritpion example. Office, Audio, Video, Games, Note... etc.

flipflop97:
The string 'Favorite' in Weblate is missing some context, especially the fact that it's a verb (and not a noun). Could that info be added?

razzeee:
I don't think that's supported by our translations. We could maybe add a screenshot, but might be hard in that specific case, as it's only used as an a11y text.

PR #1381, opened 2023-04-23
Uses boring-avatars as fallback if user avatar is null

CharlExMachina:
Addresses [this issue](https://github.com/flathub/website/issues/1366). If the avatar is null, it will use a randomly generated avatar via the `boring-avatars` package. These avatars are generated at runtime, and not being fetched from a remote source.

The end result looks like this if the user's avatar comes off as `null`:

Screenshot from 2023-04-23 15-49-51

Note that the pattern depends on the user's name; mine happens to look like that.

### DIFF
--- a/frontend/src/components/layout/Header.tsx
+++ b/frontend/src/components/layout/Header.tsx
@@ -4,7 +4,7 @@ import { useRouter } from "next/router"
 import { HiMagnifyingGlass, HiXMark, HiBars3 } from "react-icons/hi2"
 
 import { LogoJsonLd, SiteLinksSearchBoxJsonLd } from "next-seo"
-import { env } from "process"
+import Avatar from "boring-avatars"
 import { useTranslation } from "next-i18next"
 import { IS_PRODUCTION } from "../../env"
 import { useUserContext, useUserDispatch } from "../../context/user-info"
@@ -37,6 +37,7 @@ const Header = () => {
   const router = useRouter()
   const user = useUserContext()
 
+  const [userAvatarUrl, setUserAvatarUrl] = useState<string | null>(null)
   const [query, setQuery] = useState("")
 
   const [clickedLogout, setClickedLogout] = useState(false)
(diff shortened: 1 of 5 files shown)


frontend/src/components/layout/Header.tsx
```diff
@@ -248,19 +259,27 @@ const Header = () => {
...
-                                (auth) => auth.avatar,
-                              ).avatar
-                            }
-                            width="38"
-                            height="38"
-                            alt={t("user-avatar", {
-                              user: user.info.displayname,
-                            })}
-                          />
+                          {userAvatarUrl ? (
```
razzeee, line 262: This part should probably be moved to a new component, so that we don't need to repeat it (for e.g. in the other file) and keep it in sync

frontend/src/components/layout/Header.tsx
```diff
@@ -248,19 +259,27 @@ const Header = () => {
...
+                                user: user.info.displayname,
+                              })}
+                            />
+                          ) : (
+                            <Avatar
+                              size={40}
+                              name={user.info.displayname}
+                              variant="marble"
+                              colors={[
+                                "rgb(74, 144, 217)",
```
razzeee, line 278: I guess these are some of our colors (can't tell from the setup I'm on right now)

razzeee:
Thank you for working on this

frontend/src/components/layout/Header.tsx
CharlExMachina: Yes, it is. I used hardcoded values because I had issues accessing them from JS.

I'm not really sure how I would do that, as all Tailwind theme colors are used as component props and not extracted out as JS variables in any instance I could find :(

frontend/src/components/layout/Header.tsx
```diff
@@ -248,19 +259,27 @@ const Header = () => {
...
+                              src={userAvatarUrl}
+                              width="38"
+                              height="38"
+                              alt={t("user-avatar", {
+                                user: user.info.displayname,
+                              })}
+                            />
+                          ) : (
+                            <Avatar
+                              size={40}
```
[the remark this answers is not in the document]
CharlExMachina: No, it was me misreading the size above 😅, thanks for pointing it out!

frontend/src/components/layout/Header.tsx
razzeee: I don't know either, but using the hex values might at least make it easier to recognize the first one?

razzeee:
Looks great, thank you

PR #1389, opened 2023-04-24
[TRANSLATION] Add missing Turkish translations and use more known words

demirdegerli:
I added the missing translations on the footer, setup flathub button and it's description. Also I used "Popüler" instead of "Gözde" for the Popular entry on the footer. Because "Popüler" is widely used while "Gözde" does not.


razzeee:
Hey, please update via https://hosted.weblate.org/engage/flathub/ as mentioned in the readme.

I will need to close this, as merging this would break the workflow and cause us massive work. Thank you.

PR #1405, opened 2023-04-27
Add moderation frontend & backend

jameswestman:
Add an API to collect changes from flat-manager-hooks and show them to moderators and for moderators to approve or reject changes. Also added a moderation dashboard in the frontend.


frontend/src/types/Moderation.ts
```diff
@@ -0,0 +1,43 @@
+export type ModerationRequestType = "appdata"
+
+export interface ModerationAppItem {
+  appid: string
+  is_new_submission: boolean
+  updated: string
```
razzeee, line 6: Is this a date? If yous `updated_at` might be nicer?

frontend/public/locales/en/common.json
```diff
@@ -101,6 +101,26 @@
   "apps-by-projectgroup": "Apps in the {{projectgroup}} group",
   "search-for-query": "Search for '{{query}}'",
   "apps-description": "Explore the Linux apps available on Flathub",
+  "applications-description": "An app store and build service for Linux",
+  "moderation-dashboard": "Moderation Dashboard",
```
razzeee, line 105: I'm not convinced moderation should be translated. Would like to hear what others think.

razzeee: Oh I guess this new panel is not just for moderators, but for app authors too. NVM then...

frontend/public/locales/en/common.json
```diff
@@ -101,6 +101,26 @@
...
+  "moderation-new-submissions": "New Submissions",
+  "moderation-updates": "Updates",
+  "moderation-appstream-changes": "Appstream Changes",
+  "moderation-new-submission-appstream": "Appstream",
```
razzeee, line 109: This seems to be a weird key, just for `Appstream`?

backend/alembic/versions/67f5b6b8aa36_merge_moderation.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '67f5b6b8aa36'
+down_revision = ('3fba41121353', '878310d42e29')
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
razzeee, line 19: This migration does nothing?

razzeee:
Looks good so far, but I haven't checked some parts of the frontend work. Is there an easy way to have some example requests, so that I could actually look at those UIs? Maybe a csv with test data for import into postgres?

backend/alembic/versions/67f5b6b8aa36_merge_moderation.py
jameswestman: It's a merge migration, because another migration was added while I was working on this PR. I'll remove it when I make the other changes to the migration.

frontend/public/locales/en/common.json
jameswestman: App authors won't be able to see all of these strings. I can leave the ones in the moderator-only pages untranslated.

razzeee:
Checking out the frontend part, with some data entered, this seems like a great start.

Only thing I'm missing right now is a way to see some kind of audit log of what has been rejected/commented/approved. Mostly cause apps will just be gone from the queue when they are resolved. So I would guess a global page with all apps might make sense later.

backend/app/moderation.py
```diff
@@ -0,0 +1,312 @@
...
+            .filter(models.ModerationRequest.is_approved is not True)
+            .count()
+        )
+        if remaining == 0:
+            # Tell flat-manager that the job is approved
+            worker.review_check.send(request.job_id, "Passed", None)
+    else:
+        # If any request is rejected, the job is rejected
+        worker.review_check.send(
+            request.job_id, "Failed", "The review was rejected by a moderator."
```
razzeee, line 302: I'm not a native speaker, but it seems a bit weird to me, that the `review` was rejected? Wouldn't it be something else, like the upload/build/job?

frontend/src/asyncs/moderation.ts
```diff
@@ -0,0 +1,77 @@
...
+import { ModerationApp, ModerationApps } from "src/types/Moderation"
+
+/**
+ * Performs API request to retrieve list of apps for moderation
+ * @returns array of apps
+ */
+export async function getModerationApps(): Promise<ModerationApps> {
+  let res: Response
+  try {
+    res = await fetch(MODERATION_APPS_URL, { credentials: "include" })
```
razzeee, line 15: Just FYI, I might start replaceing these asyncs with something like https://tanstack.com/query/latest/docs/react/quick-start in the near future

frontend/src/components/moderation/ReviewRow.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const [status, setStatus] = useState<
+    "idle" | "pending" | "error" | "success"
+  >("idle")
+  const [error, setError] = useState<string>()
+
+  const submit = async (approve: boolean) => {
+    setStatus("pending")
+    try {
+      await submitReview(request.id, approve, comment)
+      setStatus("success")
```
razzeee, line 47: I think we might want to introduce some kind of toasts, to give the user a feedback in one of the next iterations.

backend/app/moderation.py
```diff
@@ -0,0 +1,312 @@
...
+    if new_submissions is not None:
+        query = query.having(is_new_submission == new_submissions)
+
+    return ModerationAppsResponse(
+        apps=[ModerationAppItem(**row._asdict()) for row in query]
+    )
+
+
+@router.get("/apps/{app_id}", status_code=200, response_model_exclude_none=True)
+def get_moderation_app(
```
razzeee, line 96: I think this should probably page, so that we don't run into problems, when an app has had some hundredth updates

backend/app/moderation.py
```diff
@@ -0,0 +1,312 @@
...
+    is_approved: bool | None
+    comment: str | None
+
+
+class ModerationApp(BaseModel):
+    requests: list[ModerationRequestResponse]
+
+
+@router.get("/apps", status_code=200, response_model_exclude_none=True)
+def get_moderation_apps(
```
razzeee, line 66: Should also page, even if only to keep the backend fast

backend/app/moderation.py
```diff
@@ -0,0 +1,312 @@
...
+            ),
+        )
+        .filter_by(handled_at=None, is_outdated=False)
+        .group_by(models.ModerationRequest.appid)
+    )
+
+    if new_submissions is not None:
+        query = query.having(is_new_submission == new_submissions)
+
+    return ModerationAppsResponse(
```
razzeee, line 90: I wouldn't have expected the list call to return the normal object, but a slimmed down one. With your current design, your probably only using four fields on the list page.

Then again, this might not be as bad, if we have paging and do that "early"

frontend/src/components/moderation/ReviewRow.tsx
razzeee: I just remembered, that we already have toasts :)

backend/app/moderation.py
jameswestman: This is only four fields? Unless you're talking about the other response, but we do use most of those fields.

razzeee: I guess I confused it with `ModerationRequestResponse`

frontend/public/locales/en/common.json
razzeee: Please do.

frontend/src/components/moderation/ReviewRow.tsx
jameswestman: It currently gives the error inline:

image

frontend/public/locales/en/common.json
```diff
@@ -99,6 +99,20 @@
   "apps-by-developer": "Apps by {{developer}}",
   "apps-by-projectgroup": "Apps in the {{projectgroup}} group",
   "search-for-query": "Search for '{{query}}'",
+  "apps-description": "Explore the Linux apps available on Flathub",
```
razzeee, line 102: Are you using this and the next line? Or is this from a conflict resolved wrong?

Issue #1365, opened 2023-04-22
Filter proprietary software


vanillajonathan:
Maybe there could be a way to filter out proprietary software to only show free, open source software?

In order to help users from accidentally installing proprietary software, on the app info page, if a software is proprietary software maybe it could show "Proprietary" in red color and/or with a warning icon?

Danielgomesp:
I believe we can close this issue now.

vanillajonathan:
Yeah, unless we keep it open until both PRs are merged.

PR #1408, opened 2023-04-28
Highlight proprietary license in app details page

Lesik:
Hi! This is my first PR in this repo, sorry if I did things how you usually wouldn't.

To start with something simple, I went for the second part of #1365, "if a software is proprietary software maybe it could show "Proprietary" in red color and/or with a warning icon".

I tried eyeballing the colors to find a pleasant match that draws attention yet doesn't distract too much, I think I got it alright:

Screen Shot 2023-04-28 at 17 43 54

To keep contrast between background and text according to WCAG, I also had to adjust the "subtext" (link/text) color when the ListBox is highlighted.

I was thinking of adding some sort of 🛈 icon that on hover would show a popup that explains the downsides of proprietary software because new users who don't know what proprietary means might be confused. Thoughts?

### DIFF
--- a/frontend/src/components/application/AdditionalInfo.tsx
+++ b/frontend/src/components/application/AdditionalInfo.tsx
@@ -12,6 +12,7 @@ import {
   HiInbox,
   HiLanguage,
   HiLifebuoy,
+  HiMinusCircle,
   HiScale,
   HiWrenchScrewdriver,
 } from "react-icons/hi2"
@@ -39,6 +40,9 @@ const AdditionalInfo = ({
 
   const licenseIsLink = data.project_license?.startsWith(
     "LicenseRef-proprietary=",
+  )
+  const licenseIsProprietary = data.project_license?.startsWith(
+    "LicenseRef-proprietary",
   )
 
   return (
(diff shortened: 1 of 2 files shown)


razzeee:
hey, thanks for looking into this, but I'm not sure if we want it like that. I do have some notes on how this looks etc, but I'm a bit worried about the implications of this.

@ramcq is this something we want? I know that we will always push FOSS apps to be front and center but design like this might lead to non-FOSS apps not wanting to be on flathub

razzeee:
For context gnome software:

image

The Learn More then leads here https://gitlab.gnome.org/GNOME/gnome-software/-/wikis/Software-licensing

Danielgomesp:
Hi everyone, I have designed this idea trying to address the following concerns:

- Inform that the app is proprietary;
- Avoid a design that may lead non-FOSS apps not to want to be on Flathub;
- Follow the website's current design;
- Follow the gnome-software idea

image

Or, we can just add the word "Proprietary" in front of "License"?
image


What do you guys think?

Lesik:
Thanks for the mockups @Danielgomesp! Out of these two, I personally prefer the one that has a bit of explainer text. However, I think we could take a bit more inspiration out of the GNOME Software design. Just "You may not be able to contribute to this app" doesn't really explain much, and most people might not even be  interested in contributing to random apps they're using, so they'll think "how does this affect me?". But proprietary software has more caveats than just that, as the GNOME Software tile explains. Would it be possible to incorporate a more detailed explainer to fit with the Flathub design?

razzeee:
Just cause I'm looking at permission code in gnome software right now, it also shows an app as unsafe, due to proprietary licenses.

Not sure if there is anything, that overrides it in some cases.

ramcq:
I can't really answer the design question, but I support highlighting Proprietary software and the potential downsides. I think some more textual description is needed perhaps in addition to a highlight/warning color or icon, rather than just making something red like it's broken. Flathub is after all trying to promote/support/build the free and open source Linux desktop; so we do care about favouring and promoting FOSS > Proprietary. We are just enabling developer and user choices by supporting any (distributable) software.

PR #1409, opened 2023-04-28
Filter search page for Free Software only

Lesik:
Decided to have a go at the first part of #1365, too :slightly_smiling_face: wasn't as difficult as I anticipated

This PR adds:
- `is_free_license` and `project_license` to the apps index of Meilisearch. `is_free_license` to speed up filtering ad the index can be precomputed and no string comparison is needed when the actual search is happening. `project_license` because I thought it's a useful field to have in general.
- Ability to the `/search` endpoint to filter for Free Software only. I left the default behavior at "including proprietary software" because I think that's what most people use Flathub for, getting their Spotifys, Google Chromes and Microsoft Teams in a safe way without messing up their system with half-hearted Linux packages from proprietary vendors. Edit: Not trying to imply here that's all Flathub is good for — in fact I prefer to have all my GUI apps be installed via Flatpak — just saying I have the impression it's most popular for proprietary software.
- A toggle to the search page to toggle whether to show Free Software only or not. The preference is saved to local storage for persistence. Again, I set the default to false because see bullet above. Other pages and collections are unaffected by this.

### DIFF
--- a/backend/app/apps.py
+++ b/backend/app/apps.py
@@ -1,7 +1,7 @@
 import json
 import re
-import gi
 
+import gi
 from gi.repository import AppStream
 
 from . import db, search, utils
--- a/backend/app/apps.py
+++ b/backend/app/apps.py
@@ -1,7 +1,12 @@
 import json
 import re
+import gi
+
+from gi.repository import AppStream
 
 from . import db, search, utils
+
+gi.require_version("AppStream", "1.0")
 
 
 def load_appstream():
@@ -24,6 +29,8 @@ def load_appstream():
 
             search_keywords = apps[appid].get("keywords")
 
+            project_license = apps[appid].get("project_license", "")
+
             # order of the dict is important for attritbute ranking
             search_apps.append(
                 {
@@ -31,6 +38,8 @@ def load_appstream():
                     "name": apps[appid]["name"],
                     "summary": apps[appid]["summary"],
                     "keywords": search_keywords,
+                    "project_license": project_license,
+                    "is_free_license": AppStream.license_is_free_license(project_license),
                     "app_id": appid,
                     "description": search_description,
                     "icon": apps[appid]["icon"],
--- a/backend/app/apps.py
+++ b/backend/app/apps.py
@@ -39,7 +39,9 @@ def load_appstream():
                     "summary": apps[appid]["summary"],
                     "keywords": search_keywords,
                     "project_license": project_license,
-                    "is_free_license": AppStream.license_is_free_license(project_license),
+                    "is_free_license": AppStream.license_is_free_license(
+                        project_license
+                    ),
                     "app_id": appid,
                     "description": search_description,
                     "icon": apps[appid]["icon"],
--- a/frontend/public/locales/en/common.json
+++ b/frontend/public/locales/en/common.json
@@ -293,5 +293,6 @@
   "explore-apps": "Explore apps",
   "setup-flathub": "Setup Flathub",
   "setup-flathub-description": "Setup Flathub to install and update apps on your system.",
-  "tags-colon": "Tags:"
+  "tags-colon": "Tags:",
+  "free-software-only": "Only show Free Software"
 }
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -181,8 +181,8 @@ def get_appstream(appid: str, response: Response):
 
 
 @app.get("/search/{userquery}")
-def get_search(userquery: str):
-    return search.search_apps(userquery)
+def get_search(userquery: str, free_software_only: bool = False):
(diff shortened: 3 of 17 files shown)


razzeee:
It might be nice to not do this in cookies, but in the url, then you could also send a search, to a friend, that's already prefiltered. Then again, it will probably get messy, as I'm planning to add more filters and that might lead to very long urls.

Cookie might also be nice, cause we could filter you experiance down on other pages too, but that would probably cause problems with the prebuild pages or lead to us having to prebuild everything twice.

frontend/public/locales/de/common.json
```diff
@@ -290,5 +290,6 @@
   "browse": "Durchsuchen",
   "setup-flathub": "Flathub einrichten",
   "setup-flathub-description": "Richte Flathub ein, um Anwendungen auf deinem System zu installieren und zu aktualisieren.",
-  "explore-applications": "Anwendungen entdecken"
+  "explore-applications": "Anwendungen entdecken",
```
razzeee, line 293: Please remove all translations, that arenot the `en` ones. Unfortunatly we can't touch them here, as far as I know, or it will need to conflicts - which might cause me a bunch of work :)

frontend/public/locales/en/common.json
```diff
@@ -290,5 +290,6 @@
   "browse": "Browse",
   "explore-apps": "Explore apps",
   "setup-flathub": "Setup Flathub",
-  "setup-flathub-description": "Setup Flathub to install and update apps on your system."
+  "setup-flathub-description": "Setup Flathub to install and update apps on your system.",
+  "free-software-only": "Only show Free Software"
```
razzeee, line 294: I know that @ramcq was very pro calling this floss software, but it might be hard for new users to grasp

backend/app/apps.py
```diff
@@ -24,13 +29,17 @@ def load_appstream():
...
+
             # order of the dict is important for attritbute ranking
             search_apps.append(
                 {
                     "id": utils.get_clean_app_id(appid),
                     "name": apps[appid]["name"],
                     "summary": apps[appid]["summary"],
                     "keywords": search_keywords,
+                    "project_license": project_license,
+                    "is_free_license": AppStream.license_is_free_license(project_license),
```
razzeee, line 42: I don't like how we have two terms `free license` and `free software` now. Would like them to be aligned.

razzeee:
This looks great, I only have some very minor things, listed above.

Other then this, I don't like how it flickers, when you change the toggle, as it seems to reload everything. Might be cool to just reload the items below and show a loading indicator. I'll probably look into that.

Thanks for working on this.

frontend/public/locales/en/common.json
Lesik: If it's about "free" being confused with "gratis", then perhaps "libre software" would work? I do think FLOSS can be confusing if you haven't heard the term before. To a new user, FLOSS might sound like some boring technical acronym, whereas "libre software" might sound intriguing, prompting the user to google it and learn more about it.

Another thing to consider is that in other languages, the ambiguity between "free" and "gratis" doesn't exist. To stay close to the English translation, would these languages also use the term "FLOSS" then, even though the acronym is not used in that language? And in languages not written in the roman alphabet, would it be e.g. "FLOSS програми" or "ФЛОСС програми"? Both read quite weird. 🤔

backend/app/apps.py
Lesik: I agree, I think on the search result page the term "free software" makes most sense, because you're currently trying to find software, and "Show only software with a free license" is a mouthful. I will change this field to `is_free_software`.

frontend/public/locales/de/common.json
Lesik: Removed

frontend/public/locales/en/common.json
razzeee: As far as I remember it was about the free/gratis thing and FLOSS convieniently side-stepping the libre/free discussions.

I'm not sure I have the answers for translations, I would just trust translators to do the right things, but that might also be wrong :)

razzeee:
I've pushed some changes. In general, I'm fine with it now, but let's wait for wording changes.

In the future, I'm hoping to move that filter to the side of the page and do a facet search

razzeee:
I will rebase this, please be careful, as I will force push to your branch :)

frontend/public/locales/en/common.json
razzeee: I will merge this for now, as I don't want to keep this back due to one translation.

Especially cause I will change this UI later anyway. (Facetes and stuff)

razzeee:
Thanks for working on this.

PR #1410, opened 2023-04-29
Merge landing page with apps page

razzeee:
I think this makes for a way nicer landing page and is a bit more in line with what play store does for e.g.

There is a little bit of app duplication on the page due to this now, but I don't think it's problematic

image


nikodunk:
I love this as in your screenshot alone many new apps are surfaced that're normally not on the index unless they're updated! VSCode, Gimp, Krita, LibreOffice, Blender, and many many other non-gaming apps (because let's be Popular is full of gaming :)).

razzeee:
Right now, you can see those on https://flathub.org/apps but it's way too hidden and I do think our landing page is too short.

razzeee:
@jimmac @bertob I would like to hear your opinion on this

bertob:
Side note: We really need some kind of separate category for game emulators, it's a bit nonsensical to have a "games" category where the 6 top games are all emulators and not actual playable games.

razzeee:
While I'm very unhappy with the freedesktop categories, I would try to go without merging for now, but I agree, that we shourd prioritize them. (how ever I should figure out the order)

While a quality gate would be good, it only really makes sense if it isn't a one way street. So if we ban something from the landing page, the dev should be informed to as why. This will only become possible after moderation #1405  I guess, if even then.

Regarding emulators https://github.com/flathub/website/pull/1240
But according to spec it's just a subcategory of games (and system) and those won't show on the landing page. And as they are subcategories, we can't/shouldn't substract them from the games category I guess?

bertob:
> While a quality gate would be good, it only really makes sense if it isn't a one way street. So if we ban something from the landing page, the dev should be informed to as why. 

Yeah, agreed. Ideally we'd expose the rules clearly so people already fix these issues before even submitting their apps to Flathub, but we'd definitely need some kind of structure that greenlights what goes onto the home page, at least. What I was imagining is a curation/admin panel that shows new apps and ones that recently changed their metadata (so that apps that fix their metadata show up), and allows the curation team to allow/denylist those apps.

Allan and I started drafting some metadata quality guidelines a while back, need to pick that back up.

PR #1414, opened 2023-04-29
fix: hamburger menu closes #1413

### LINKED ISSUE #1413: Open hamburger menu messes up large website view
When leaving the hamburger menu open and going from the small mobile view into the larger view (through the developer mobile tools on Firefox), the website's content is gone. It could be that this really doesn't matter as few people will ever encounter this, but just thought I'd create an issue for it which can be closed in case this isn't worth fixing. The only real life scenario where I could see this happen is on a mobile device with a significant difference between its landscape and portrait layout.

## How to reproduce
1. Open the mobile developer tool in the browser (ctrl+shift+m on Firefox)
2. Make the screen small enough to trigger the smallest version of the website
3. Open the hamburger menu
4. With the hamburger menu open, drag the site to the largest version again

The content of the site is now gone.

umairraza96:
Add `useWindowSize` hook to determine the window size which is required to fix this issue

This PR closes issue #1413


frontend/src/components/layout/Header.tsx
```diff
@@ -124,6 +124,12 @@ const Header = () => {
     }
   }, [])
 
+  // function toClose() {
```
razzeee, line 127: This seems unintentional

umairraza96: Yes it is unintentional, I have removed it though thanks for noticing :+1:

frontend/src/components/layout/Header.tsx
```diff
@@ -331,7 +333,12 @@ const Header = () => {
               </div>
             </div>
 
-            <Popover.Panel as="nav" className="lg:hidden" aria-label="Global">
+            <Popover.Panel
+              as="nav"
+              className="lg:hidden"
+              aria-label="Global"
+              ref={popoverRef}
```
razzeee, line 340: The ref doesn't seem to be needed? Can we remove it?

umairraza96: got it, thanks for letting me know! :+1:

razzeee:
Thanks for looking into this

PR #1436, opened 2023-05-01
feat: error boundary closes #456

### LINKED ISSUE #456: Implement error boundaries
Seems to be best practice to have these?
https://reactjs.org/docs/error-boundaries.html

umairraza96:
Right now I am displaying `Custom404` component if  I could get UI for custom error I can easily change it

closes #456

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -17,6 +17,9 @@ import { bcpToPosixLocale, languages } from "../src/localize"
 import Main from "../src/components/layout/Main"
 
 import { Inter } from "next/font/google"
+import * as Sentry from "@sentry/react"
+import { Error } from "../src/components/Error"
+
 const inter = Inter({
   subsets: ["latin"],
 })
@@ -71,4 +74,6 @@ const App = ({ Component, pageProps }: AppProps) => {
   )
 }
 
-export default appWithTranslation(App)
+export default Sentry.withErrorBoundary(appWithTranslation(App), {
+  fallback: <Error />,
+})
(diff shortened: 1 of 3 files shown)


frontend/pages/_app.tsx
```diff
@@ -62,7 +63,9 @@ const App = ({ Component, pageProps }: AppProps) => {
         />
         <UserInfoProvider>
           <Main className={inter.className}>
-            <Component {...pageProps} />
+            <ErrorBoundary>
```
razzeee, line 66: Could it make sense to move this up a few levels or should it be next to the component? `UserInfoProvider` could also error I guess?

Or would that not work?

razzeee: Nevermind, I just started reading up on it

frontend/src/components/layout/ErrorBoundary.tsx
```diff
@@ -0,0 +1,40 @@
...
+  static getDerivedStateFromError() {
+    return { hasError: true }
+  }
+
+  render(): React.ReactNode {
+    const { hasError } = this.state
+
+    // render fallback ui
+    if (hasError) {
+      return <Custom404 />
```
razzeee, line 33: I think it shouldn't be a 404, maybe use the layout from a 404 page and say that there was an error and maybe a retry button like the docs have? https://nextjs.org/docs/advanced-features/error-handling

razzeee:
Thanks for looking into this, looks great, we only need to figure out the UI we want to show when this happens :)

razzeee:
Added a new error page, this should be good to go. But we will need to keep an eye on sentry, as I'm not sure if this might keep it from reporting errors.

razzeee:
I've got an answer over in the sentry repo, we need to handle the sentry part before we merge this.

razzeee:
Completly reworked according to upstream. I took the liberty to squash this all, as all code changed.

PR #1437, opened 2023-05-01
Add distro setup instructions

razzeee:
This might not be something we can actually merge, but let me explain.

This grabs the flatpak website repo at build and copies over all images and the infos for distros/setups.

Then it renders them.

Cons: 
- We heavily depend on a different repo
- Some of the pages are a bit wild
- No translations

image

image

image

PopOS probably looks the worst:

image


JakobDev:
Why do we need this? If this has the exact same Content as https://flatpak.org/setup/, we not just link there?

razzeee:
That's what we do right now.

There was talks about improving it, as it's hard to find distros over there. So this allows you to search and we sort some distros to the top.

This would also be the first step to be able to detect your distro (if the browser permits) and automatically show you the right instructions.

So right now we fear to loose newer users e.g. on Ubuntu in the process

PR #1449, opened 2023-05-01
Add a permission tile

razzeee:
Heavily based on gnome softwares implementation. Only the upper level without the modal for now.

I haven't found a place, where it differs from what gnome software thinks, but I'm sure there will be places.

UI needs finetuning.

image

image

image


frontend/styles/main.scss
```diff
@@ -51,6 +51,10 @@
   --flathub-celestial-blue: 74 144 217;
   --flathub-electric-red: 226 0 0;
 
+  --flathub-status-green: 67 160 71;
```
razzeee, line 54: @JakobDev could you come up with some status colors? I went for some muted ones, but you will probably know better (and the red is just the one we had before)

frontend/src/safety.ts
```diff
@@ -0,0 +1,353 @@
...
+    )
+  ) {
+    appSafetyRating = addToSafteyRating(
+      appSafetyRating,
+      SafetyRating.potentially_unsafe,
+      "can-access-and-change-user-settings",
+    )
+    noPermissions = false
+  }
+
```
TingPing, line 287: Suggestion:

`org.freedesktop.secrets` -> "Can read user passwords and keys stored on the system"

TingPing, line 287: For all not listed dbus access, I think we have to consider it potentially unsafe and list them out.

We could potentially describe all `--own-name=` permissions as ability to impersonate other software?

TingPing: I think we can whitelist `org.mpris.MediaPlayer2.*`

frontend/styles/main.scss
JakobDev: Sorry to disappoint you, but I'm not really good at designing, so I can't really help you with share of green would look the best here.

razzeee: sorry @JakobDev I meant to ping @jimmac  :)

jimmac: I don't see why not embrace the low hanging fruit -- just do what libadwaita does. Sadly light/dark situation is throwing a wrench in the simple approach. CSS doesn't yet have a reasonable way to do some SASS-styled color mixing, so I've seen people using HSL definitions using custom properties for the individual components and then using calc() to tint the lightness for the dark variant. But even libadwaita just hardcodes the light/dark variants:

```
/* libadwaita colors */
$success: #2ec27e;
$success-dark: #26a269;
$warning: #e5a50a;
$warning-dark: #cd9309;
$error: #e01b24;
$error-dark: #c01c28;
```

turbervillejoseph (approved):

orowith2os:
Showing a list of permissions would be far too technical for the end user, and imo a GUI that presents it like GNOME Software does, does it best. 

Most common cases can be covered here, and it appears that they are - if something is iffy, it gets noted in an easily accessible way.

razzeee:
Detail modal

image

orowith2os:
I'd recommend to changing the verification bit there to "This distribution is verified" - it's possible the software developer isn't, and the package is, or the topic of the app gets it in a bit of a sticky situation.

razzeee:
> I'd recommend to changing the verification bit there to "This distribution is verified" - it's possible the software developer isn't, and the package is, or the topic of the app gets it in a bit of a sticky situation.

Wording is taken from gnome software, but they might mean something different

frontend/src/safety.ts
razzeee: Can you raise this in https://gitlab.gnome.org/GNOME/gnome-software/-/issues (and maybe ping me)

razzeee:
As this has been hidden behind a feature flag, I'll likely merge this in the coming days - it won't be visible to users.

PR #1454, opened 2023-05-04
Remove project groups

razzeee:
The grouping is too arbitrary, especially as we try to increase user trust.


razzeee:
This is missing redirects for the old pages. I would just let them 404 - google will probably understand.

JakobDev:
> especially as we try to increase user trust.

Why does this hurt the user trust?

razzeee:
Cause everybody can just change their appstream file and slide into different groups. This would change with moderation.

Showing the unofficial steam flatpak under the official steam link flatpak is problematic
https://flathub.org/apps/com.valvesoftware.SteamLink

Additionally KDE and GNOME groups specifically look way more official to a normal user then they are.

JakobDev:
> Cause everybody can just change their appstream file and slide into different groups. This would change with moderation.

Flathub is currently moderated. You can't just upload any App to Flathub. You need to create a PR with the full Manifest which will be reviewed before publishing.

> Additionally KDE and GNOME groups specifically look way more official to a normal user then they are.

Fair Point, but the KDE Group is currently the only Way to see all Apps of KDE

> The developer grouping will also have to change.

What you will do to see which Apps belongs to a Developer if not using AppStream? Also, is this really needed? Is there any example of a App on Flathub who says it's e.g. official from KDE/Gnome but it is not?

razzeee:
> Flathub is currently moderated. You can't just upload any App to Flathub. You need to create a PR with the full Manifest which will be reviewed before publishing.

Yes, but you can then just change that in the repo and no moderator will notice.

> Fair Point, but the KDE Group is currently the only Way to see all Apps of KDE

There is https://flathub.org/apps/search/org%2Ekde which is something

> What you will do to see which Apps belongs to a Developer if not using AppStream? 

The idea is to connect this to the appid/verification I guess. But that idea is pretty fresh.

> Also, is this really needed? Is there any example of a App on Flathub who says it's e.g. official from KDE/Gnome but it is not?

We for example have the official steam link linking to the unofficial steam - which is a bit confusing

JakobDev:
> Yes, but you can then just change that in the repo and no moderator will notice.

If that the only Problem, it can be pretty easy solved. Ypu need to pass all the checks of the Buildbot before allowed to merge. Give me a few Minutes and I can write you a script, that checks if the Developer Name in the AppStream the same as on Flathub and if not, it fails.

> There is https://flathub.org/apps/search/org%2Ekde which is something

It is possible with the search function, but there should be a better Way.

> The idea is to connect this to the appid/verification I guess. But that idea is pretty fresh.

I don't think that will works well. There are People who using a weird domain that is not their own Name. Also there are People on GitHub who wants to show their real Name and not their GitHub Username.

> We for example have the official steam link linking to the unofficial steam - which is a bit confusing

Why is that a problem? Both are from the same Publisher, so they should be shown together. The fact that one is official (what is clearly shown) and one not does not matter. Also Valve (I'm 100% sure they know about the Steam Flatpak) don't seem to have a Problem with this.

barthalion:
The grouping should be done using ID instead of using an arbitrary appdata field. There's not much to discuss here; either someone steps up to make it work properly or it should be dropped.

PR #1468, opened 2023-05-08
Don't allow logos outside our hosted ones

razzeee:


JakobDev:
Is there maybe a Way to mirror remote Icons on Build (same as Screenshots), so Apps which using this are not missing their Icon?

razzeee:
Some apps are using remote icons and are copying the files in their install. That seems to also tip of appstream, that there are other icon sources, which seem to get used instead.

PR #1516, opened 2023-05-23
Support for creating direct upload apps

jameswestman:
Flathub hopes to support apps that are directly uploaded by an upstream project (e.g. by their CI pipeline) rather than going through Buildbot and the repositories in https://github.com/flathub. To support that, we want developers to be able to create apps in the backend rather than through GitHub. In the future, they will be able to create an upload token for flat-manager.

This PR adds support for verifying apps that don't have a https://github.com/flathub repository. (If the repository for an app ID does exist, you still must have access to it). Verifying the app adds it to the new BackendApp table, which is used as a source for dev_flatpaks.

There is a "New App" button on the developer page, which leads to a page describing the difference between "community repository" and "direct upload" apps. The "direct upload" button allows you to enter an app ID and verify it, and completing verification brings you to the app's developer page.

Creating upload tokens for a backend app is probably best left to another PR.


JakobDev:
Does that mean a prebuilt App can be uploaded without using Buildbot to build or does that mean, it is possible to give Buildbot the Manifest directly without using GitHub?

razzeee:
As far as I know, buildbot or a replacement will always be an option. That's at least the plan I know of for now.

JakobDev:
I hope that Buildbot will stay. For me it's more comfortable to just edit the manifest on GitHub than maintaining my own CI Pipeline. It is also needed for Flatpaks, which are not maintained by upstream. But the Problem still is, that we will likely loose ARM Builds for most Flatpaks who use this new Method.

razzeee:
As said, buildbot will stay or be replaced by something else, that can build stuff and give us more/modern features. But that's pretty far away.

JakobDev:
The Problem here is, that if you use the new Feature of this PR, Buildbot is not involved.

razzeee:
That is precisely the point - to give more options. Right now only OBS and Firefox can do this and whenever the token expires, they need to find barth and get a new one from him, that doesn't scale.

razzeee:
If you move from buildbot to your own infra, that could happen, but if that's the case, you probably did not support arm before.

JakobDev:
Not really. There are many Apps which works perfectly on ARM, but don't provide ARM binaries.

ramcq:
We're hoping to make an alternative to Buildbot over time, but as @razzeee says, Buildbot isn't going anywhere - the point is not to get people off our infrastructure, it's to enable people with their own existing infrastructure, or build systems that work badly with flatpak-builder, to upload to Flathub more easily.

backend/alembic/versions/967305fbe7aa_backend_app.py
```diff
@@ -0,0 +1,38 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('backendapp',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('app_id', sa.String(), nullable=False),
+    sa.Column('owner', sa.Integer(), nullable=False),
```
razzeee, line 24: This will cause users with apps not being able to delete their accounts. We probably need a workflow for that. I don't think we want to remove the app in that case?

razzeee, line 24: I think this should not be 1:n but n:n as I think it should be possible to have multiple owners.

backend/app/verification.py
```diff
@@ -233,6 +237,24 @@ class VerificationStatus(BaseModel):
...
+def _is_github_app(appid: str) -> bool:
+    """Determines whether the app is a repo in github.com/flathub."""
+    try:
+        gh = github.Github(
+            config.settings.github_client_id, config.settings.github_client_secret
+        )
+        try:
+            gh.get_repo(f"flathub/{appid}")
+        except UnknownObjectException:
+            return False
```
razzeee, line 249: This is potentially unsafe? I would prefer to return `True` here

backend/app/verification.py
```diff
@@ -233,6 +237,24 @@ class VerificationStatus(BaseModel):
...
 
 
+def _is_github_app(appid: str) -> bool:
+    """Determines whether the app is a repo in github.com/flathub."""
+    try:
+        gh = github.Github(
+            config.settings.github_client_id, config.settings.github_client_secret
+        )
+        try:
+            gh.get_repo(f"flathub/{appid}")
```
razzeee, line 247: Wondering, if we need to check `archived` here. And how the flow would work, to move from a github repo to our own, but keeping the appId the same.

razzeee: Checking and discarding `archived`, would at least allow apps, that already moved like firefox / obs to create their account. Still we somehow need to prevent others from doing that before the right person does it.

backend/app/verification.py
```diff
@@ -285,8 +307,25 @@ def _check_app_id(appid: str, login):
...
+                status_code=403, detail=ErrorDetail.MUST_ACCEPT_PUBLISHER_AGREEMENT
+            )
+
+        if models.BackendApp.by_app_id(sqldb, appid) is not None:
+            raise HTTPException(status_code=400, detail=ErrorDetail.APP_ALREADY_EXISTS)
+
+        if _is_github_app(appid):
+            raise HTTPException(status_code=400, detail=ErrorDetail.APP_ALREADY_EXISTS)
+
+        if _matches_prefixes(appid, "com.github", "com.gitlab"):
```
razzeee, line 322: This means you can't migrate your grandfathered app, with those ids, unless you rebase onto a new id. Which might be fine.

frontend/public/locales/en/common.json
```diff
@@ -332,5 +332,18 @@
...
+  "direct-upload-description": "Build your app in your own continuous integration environment and upload it directly to Flathub.",
+  "new-direct-upload": "New Direct Upload App",
+  "verify-app-id": "Verify App ID",
+  "app-id": "App ID",
+  "app-id-instructions": "App IDs use reverse domain name notation (RDNN). For example, GNOME Maps uses 'org.gnome.Maps'. Pick a domain you have control over, such as the project's website (or, if you're uploading as an individual, your own website). If no website is available, you can also use a domain associated with a source code hosting account, such as 'io.github.<your_username>.<app_name>'.",
+  "app-already-exists": "An app with this ID already exists.",
+  "malformed-app-id": "The app ID is malformed. App IDs must contain at least three segments, and no segment may be empty. Each segment must start with a letter or underscore and may only contain letters, digits, underscores, and hyphens.",
+  "publisher-agreement": "Publisher Submission Agreement",
+  "publisher-agreement-accept": "I accept the Publisher Submission Agreement"
 }
```
razzeee, line 349: String for `new-app` is missing

frontend/pages/apps/new/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+        <h2 className="mb-6 mt-12 text-2xl font-bold">
+          {t("community-repository")}
+        </h2>
+        <p>{t("community-repository-description")}</p>
+        <ButtonLink
+          passHref
+          href="https://docs.flathub.org/docs/for-app-authors/submission#how-to-submit-an-app"
+          target="_blank"
+        >
```
razzeee, line 25: Add `rel="noreferrer"` even if it's our page (and modern browsers do it)

razzeee:
Great work, got some minor things above and some small thoughts.

1. Doing this in the frontend, it leaves me with my new backend app and being able to verify it, but no further action? Showing the secrets is the next step I guess?
2. I would like to do another frontend pass after this is merged. It would be nice to make the texts a bit nicer, before that, to not cause multiple translation passes. I mostly mean text formatting.

backend/alembic/versions/967305fbe7aa_backend_app.py
jameswestman: I was planning to have a separate table for that (in a later PR). There would be one owner, who could invite other developers. But that could be something to discuss.

jameswestman:
Generating upload tokens is the next step, yeah.

backend/alembic/versions/967305fbe7aa_backend_app.py
razzeee: Makes sense, I guess the real question would be, if this will bite us later on / we would need to change it.

Worst I can think of, is admins having to move owners around, as people become unresponsive. And maybe people complaining that there can only be one owner, as they are used to having multiple from github/gitlab I guess.

jameswestman:
I added a separate table for developer access, so it can be n:n instead of 1:n. It has an is_primary column so we can track who the "primary developer" who originally registered the app is. In the future they'll be able to invite more developers or transfer their primary status.

backend/app/verification.py
jameswestman: Wouldn't that also allow people to revive an EOL'd app? Is that something we want to prevent?

backend/alembic/versions/967305fbe7aa_backend_app.py
jameswestman: I went ahead and added the extra table (but not the invite system) to this PR.

jameswestman: I'd assume we don't want an app to have no owner, and that if you're the last developer then you have to transfer ownership to a new maintainer or EOL the app before you can delete your account.

backend/app/verification.py
razzeee: I think that would be fine, they verified the app first. So unless they somehow got hold of that url/git account, they should be able to revive an app.

I think we allow apps to be revived right now too, if there is a new maintainer, that takes over the flathub github repo.

razzeee:
This probably closes https://github.com/flathub/website/issues/230

PR #1535, opened 2023-06-01
backend: Add email system

jameswestman:
Add a worker for sending emails to an individual user or to the developers of an app.

Currently, an email is sent when someone logs in, when a build is held for moderation, and when a moderator reviews a change.

Screenshot from 2023-06-01 17-32-35

### DIFF
--- a/backend/app/email_templates/base.html
+++ b/backend/app/email_templates/base.html
@@ -1,0 +1,45 @@
+<div class="header">
+  <img src="https://flathub.org/img/logo/flathub-logo-toolbar.svg" alt="Flathub"/>
+</div>
+
+<hr/>
+
+<h1>{{ email_subject }}</h1>
+
+{% block content %}
+{% endblock %}
+
+<hr/>
+
+{% if app_id %}
+  You are receiving this email because you are a maintainer of the app {% include "app_name_and_id.html" %} on Flathub.
+{% else %}
+  You are receiving this email because you have an account on Flathub.
+{% endif %}
+
+
+<style>
+  .header {
+    display: flex;
+    justify-content: space-around;
+    margin: 2em;
+  }
+  .header img {
+    height: 2.5em;
+  }
+  h1 {
+    font-size: 1.5em;
+  }
+  h2 {
+    font-size: 1.3em;
+  }
+  hr {
+    margin-top: 1.5em;
+    margin-bottom: 1.5em;
+  }
+
+  th, td {
+    text-align: start;
+    padding-right: 1em;
+  }
+</style>
--- a/backend/app/email_templates/base.html
+++ b/backend/app/email_templates/base.html
@@ -1,3 +1,9 @@
+{% if env != "production" %}
+  <p style="border: 1px solid black; border-radius: 0.25em; padding: 0.5em;">
+    This email was sent from a {{ env }} instance.
+  </p>
+{% endif %}
+
 <div class="header">
   <img src="https://flathub.org/img/logo/flathub-logo-toolbar.svg" alt="Flathub"/>
 </div>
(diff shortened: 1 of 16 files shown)


backend/app/email_templates/base.html
```diff
@@ -0,0 +1,45 @@
...
+{% endblock %}
+
+<hr/>
+
+{% if app_id %}
+  You are receiving this email because you are a maintainer of the app {% include "app_name_and_id.html" %} on Flathub.
+{% else %}
+  You are receiving this email because you have an account on Flathub.
+{% endif %}
+
```
razzeee, line 19: No way to opt out? Not sure if we would need that

backend/app/email_templates/security_login.html
```diff
@@ -0,0 +1,7 @@
+{% extends "base.html" %}
+
+{% block content %}
+
+Someone recently logged into your account on Flathub using the <strong>{{ provider }}</strong> account <strong>{{ login }}</strong>. If it wasn't you, please contact us immediately.
```
razzeee, line 5: Wondering if we should say how? Not sure the mail address this mail is coming from is going to be one that's getting read? @barthalion

backend/app/emails.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if gl_user := models.KdeAccount.by_user(db, user):
+        return _get_gitlab_email(gl_user, "https://invent.kde.org")
+
+
+def _create_message(
+    user: models.FlathubUser, info: EmailInfo, db
+) -> tuple[str, MIMEText]:
+    email = _get_email_address(user, db)
+    full_subject = info.subject
```
razzeee, line 66: Having written an email system at work this week too, I would like to see us prefix the subject, if the mail is coming from a test system. (which we don't have right now)

I tihkn you should be able to use the `ENV` env var for that?

backend/app/emails.py
```diff
@@ -0,0 +1,135 @@
...
+    for dest, message in messages:
+        # Queue each message separately so that if one fails, the others won't be resent when the task is retried
+        worker.send_one_email.send(message.as_string(), dest)
+
+
+def send_one_email(message: str, dest: str):
+    if not settings.smtp_host:
+        print(f"Would send email to {dest}:\n{message}\n")
+    else:
+        with SMTP(settings.smtp_host, settings.smtp_port) as smtp:
```
razzeee, line 133: Wondering, if we should also log this to a database table so that we can audit/debug this, as I have seen systems doing that.

Then again, the less data we have the better :)

razzeee (approved):
Some minor things to talk about

backend/app/email_templates/base.html
jameswestman: I'm planning to add that at some point. It'll need a backend for storing preferences and a frontend for editing them, but that shouldn't be too difficult.

I think there's some emails we don't want people to opt out of, like when a token is generated. It might also be a good idea to notify people if a notification category is disabled.

backend/app/emails.py
jameswestman: That's a good idea. We don't currently set ENV in the docker-compose.yml though, I suppose we should.

razzeee: Keep in mind, that the docker compose is only for development and isn't used on the server. It should still get the env.

backend/app/email_templates/base.html
razzeee: I think we agreeded yesterday, that we won't let people opt out, as it's central funtionallity.

backend/app/emails.py
razzeee: This has since been added

PR #1556, opened 2023-06-06
Don't show localized Screenshots

JakobDev:
Localized  are possible since the libappstream port. I already use it in [one of my Apps](https://flathub.org/apps/page.codeberg.JakobDev.jdReplace), which confuses the Frontend. As translated AppStream Data is not wanted on the Website, this PR hides localized Screenshots.


razzeee:
Thank you!

PR #1585, opened 2023-06-15
storefront-info: Add is-free-software endpoint

jameswestman:
flat-manager-hooks needs to know whether an app is free software based on the license in the uploaded appstream file, not the old version that the backend has (or might not have, if the app is new). The logic for deciding if an app is FOSS is complicated, so instead of duplicating it in the hook, I added an endpoint that decides based only on the app ID and license.


razzeee:
A test call would be nice, but I'll merge it for now.

PR #1586, opened 2023-06-16
feat: i18n ide extension and machine translation CLI

NilsJacobsen:
> Let me know if the PR needs any changes.

## Feature Preview

Flathub already uses `weblate` as the i18n editor solution to contribute translations. In addition to that I highly recommend adding the ide extention from `inlang`. Already contributed to inlang and can tell that the community is bringing better DX to the i18n process week for week.

Docs -> https://inlang.com/documentation/apps/ide-extension

I added a config, which is the only thing to do. It is build for i18next so it will work perfectly for you. This allows you to use the [IDE extension](https://inlang.com/documentation/apps/ide-extension) for `vscode` with no additional configuration. It's a simple helper to extract messages and have inline annotations. 

You also get ootb the [inlangCLI](https://inlang.com/documentation/apps/inlang-cli) to lint & machine translate the i18n strings. Just installed it because it's a quick win for the devs who want to use it and doesn't need any more code or integration efforts.

I also added the extension to the workspace recommendations so everybody can use it right away and benefit from this shortcut.

#### Extract Messages (translations)

Extract Messages (translations) via the `Inlang: Extract Message` code action.

Extract Messages

#### Inline Annotations

See translations directly in your code. No more back-and-forth looking into the translation files themselves.

Inline Annotations


inlang.config.js
```diff
@@ -0,0 +1,18 @@
+export async function defineConfig(env) {
+    const { default: i18nextPlugin } = await env.$import(
+      "https://cdn.jsdelivr.net/npm/@inlang/plugin-i18next@2/dist/index.js"
```
razzeee, line 3: Can this be done in a way, that it would also work offline? / Doesn't change without us knowing?

inlang.config.js
```diff
@@ -0,0 +1,18 @@
+export async function defineConfig(env) {
```
razzeee, line 1: This file is probably not in the correct folder?

inlang.config.js
NilsJacobsen: I could add the plugin to the repo. Just keep in mind that you can access updates automatically.

inlang.config.js
NilsJacobsen: Inlang maybe will support that in the future too, but for now the config file needs to be in the root dir.

inlang.config.js
NilsJacobsen: Ah I thought that you want it completely isolated in the repo. Added the package.

inlang.config.js
```diff
@@ -0,0 +1,18 @@
...
+    const { default: i18nextPlugin } = await env.$import(
+      "./frontend/node_modules/@inlang/plugin-i18next/dist/index.js"
+    );
+    const { default: standardLintRules } = await env.$import(
+      "https://cdn.jsdelivr.net/npm/@inlang/plugin-standard-lint-rules@3/dist/index.js"
```
razzeee, line 6: This then also needs to be moved over

NilsJacobsen: Sorry, u are right :) Done ✅

razzeee:
I don't think we want this. Had some time to play with this today.

The editor command complains about an `unsupported platform`.

Lint seems to complain about every translation, which is probably due to our workflow and could be changed via linting rules, but the question is if there is much to get. Our translators are not in the codebase after all.

Showing the translation next to the key is nice, but most times it's just the key without dashes in between.

I also don't think we want machine translations, as it might keep people from doing a regular translation.

PR #1598, opened 2023-06-22
Replace Poetry with pip-tools

barthalion:


razzeee:
Why?

Please don't :)

barthalion:
Why not? Poetry does seem overcomplicated, pip-tools is much easier to reason about.

razzeee:
Well for me this complicates things. I understand how poetry does stuff, as it's pretty similar to npm. I never understood requirement files/setup.cfg and I have given up all hope to understand them.

backend/pyproject.toml
```diff
@@ -1,48 +1,5 @@
...
-requests = "^2.31.0"
-redis = "^4.5.5"
-feedgen = "^0.9.0"
-PyGObject = "^3.44.1"
-watchgod = "^0.8"
-FastAPI-SQLAlchemy = "^0.2.1"
-alembic = "^1.11.1"
-psycopg2-binary = "^2.9.6"
-itsdangerous = "^2.1"
-PyGithub = "^1.58"
```
razzeee, line 23: It also seems way easier to be able to figure out which version is installed/targeted. With the new tooling, you would need to look into the requirement file, to even get a rough version.

I guess there would be a cli command, but that doesn't help you if you're looking at it on github.

barthalion: I mean, you can just look into requirements.txt to see what it has been resolved to. It doesn't functionally change anything, apart from the dependences being defined in requirements.in. Managing deps in pyproject.toml is also supported.

PR #1661, opened 2023-07-08
index.tsx: add line break to "search element not found" page

kbdharun:
## Context

Currently, when I search for an application which isn't present the below text is displayed

image

The "Could not find a match for search." text is placed close to the heading, making it look cramped in some translations and small device form factors.

image

## Change

This PR adds a line break to give space between the heading and subsequent text.

image

image


razzeee:
I'll have a look later IMO, having a linebreak isn't the best approach

razzeee:
Closing this in favor of my PR, thank you

PR #1675, opened 2023-07-11
Allow the backend to parse more summaries

razzeee:
This enables parsing the summaries of console applications like wine


razzeee:
Lol, checking this again, it seems like Wine is the only app that had this `bug` in it's meta data parsing. We'll let's fix it.

PR #1676, opened 2023-07-11
Track app type and add console apps to search index

### LINKED ISSUE #53: support non-desktop components
Currently, the api endpoints are desktop-app specific. We should expose at least cli, extensions & sdk's. This also would be a step forward to support displaying extensions of an application.

razzeee:
Closes #53


razzeee:
Console apps will only show up when they have both icon and screenshots.

But there seem to be some with broken screenshots, probably due to that incident in the early parts of the year:

cc.nift.nsm
com.openwall.John
com.github.ilai_deutel.kibi
radio.k0swe.Kel_Agent

PR #1678, opened 2023-07-12
Add arches to search

razzeee:


razzeee:
image

This kind of raises the question again, if we can communicate the arches better somehow.

orowith2os:
> This kind of raises the question again, if we can communicate the arches better somehow.

I don't think so, that's the best we can do. Anything else can be inaccurate.

Also good to see so many apps being available for aarch64 on Flathub!

PR #1684, opened 2023-07-14
Adding Test in frontend components 

Claudio-code:
Issue https://github.com/flathub/website/issues/1184

### Flows to be tested

- [x] Login redirect component
- [x]  Logout Button in header


frontend/__tests__/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <>
+        <UserContext.Provider value={userState}>
+          <UserDispatchContext.Provider value={dispatchMock}>
+            <Header />
+          </UserDispatchContext.Provider>
+        </UserContext.Provider>
+      </>,
+    )
+
+    await waitFor(() => {
```
razzeee, line 58: We could also check for the avatar icon with user name here

frontend/__tests__/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      fireEvent.click(getByText("open-user-menu"))
+      fireEvent.click(getByText("log-out"))
+    })
+
+    expect(dispatchMock).toHaveBeenCalledWith({ type: "loading" })
+    expect(dispatchMock).toHaveBeenCalledWith({ type: "logout" })
```
razzeee, line 64: Check if we have the login button now?

frontend/__tests__/components/login/ProviderLink.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  method: "github",
+  name: "GitHub",
+}
+const githubProps: Props = {
+  inACard: true,
+  provider: githubProvider,
+}
+const translationMock = jest.fn()
+
+jest.mock("react-i18next", () => ({
```
razzeee, line 22: We probably want to mock this globally

frontend/__tests__/components/login/ProviderLink.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+      i18n: {
+        changeLanguage: () => new Promise(() => {}),
+      },
+    }
+  },
+}))
+
+const useStateSpy = jest.spyOn(React, "useState")
+const setStateMock = jest.fn()
+// @ts-ignore
```
razzeee, line 35: Why do we need this?

frontend/__tests__/components/login/ProviderLink.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        provider={githubProps.provider}
+        inACard={githubProps.inACard}
+      />,
+    )
+
+    await waitFor(() => {
+      fireEvent.click(getByRole("button"))
+    })
+    expect(setStateMock.mock.calls).toEqual([[true], [false]])
+    expect(toast.error).toHaveBeenCalled()
```
razzeee, line 128: These expectations would also pass on the previous test, so that's probably an anti pattern. Not sure if in the test or in the components behavior.

frontend/__tests__/components/login/ProviderLink.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  })
+
+  it("redirect login test if clicked is true not call api", async () => {
+    global.fetch = jest.fn() as jest.Mock
+    // @ts-ignore
```
razzeee, line 133: Why is this needed?

razzeee:
Great work, other then the mentioned comments, I think this misses one crucial thing. It needs to be added to the github pipeline. Maybe as a new pipeline step, so that we can validate these pass on every merge.

Claudio-code:
Cool i make adjustments during the week

- [ ] We could also check for the avatar icon with user name here
- [ ] Check the error test on the component
- [ ] Add mock global to react-i18next
- [ ] Remove all @ts-ignore
- [ ] In logout test add check if have login button
- [ ] Add tests in github pipeline

frontend/__tests__/components/login/ProviderLink.test.tsx
Claudio-code: was removed

frontend/__tests__/components/login/ProviderLink.test.tsx
Claudio-code: was removed

frontend/__tests__/components/login/ProviderLink.test.tsx
Claudio-code: changed to global mock

frontend/__tests__/components/login/ProviderLink.test.tsx
Claudio-code: I agree, I'll just leave the validation of the toast text, which is the only difference between the error scenarios

frontend/__tests__/components/layout/Header.test.tsx
Claudio-code: cool added it test too

frontend/__tests__/components/layout/Header.test.tsx
Claudio-code: cool added it test

Claudio-code:
@razzeee I made the adjustments if you can review again

.github/workflows/frontend_ci.yml
```diff
@@ -43,6 +43,9 @@ jobs:
       - name: Lint
         run: yarn lint
 
+      - name: Unitary Tests
```
razzeee, line 46: I've never seen it called this either `Unit tests` or better just `Frontend tests`

frontend/__tests__/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <>
+        <UserContext.Provider value={userState}>
+          <UserDispatchContext.Provider value={dispatchMock}>
+            <Header />
+          </UserDispatchContext.Provider>
+        </UserContext.Provider>
+      </>,
+    )
+
+    const userIconBeforeLogout = container.querySelector("img.rounded-full")
```
razzeee, line 55: I would much rather see us using testing library here to also ensure accessability

https://testing-library.com/docs/queries/about#priority

.github/workflows/frontend_ci.yml
Claudio-code: Changed stage name

frontend/__tests__/components/layout/Header.test.tsx
Claudio-code: changed to getByRole

razzeee:
Hrm, seems like it's not running on our CI

Claudio-code:
there was a conflict in the yarn.lock file

razzeee:
I guess we're running into https://github.com/vercel/next.js/issues/40183

Claudio-code:
I'm trying to apply the issue solution, but it's hard to find packages that use esm

razzeee:
It also makes it sound like it's an regression on their end, so maybe if they fix it upstream we could wait for it.
I guess we should test if the downgrade like someone suggests in the issue helps in our case.

Claudio-code:
I tried downgrading jest to 27 and next to 13.2 and 13.1 don't work, but I  if you remove node_modules and yarn.lock works but if you remove just remove node_modules not works,  I don't know why

PR #1714, opened 2023-07-29
common.json: fix typos, minor improvements

kbdharun:


frontend/public/locales/en/common.json
```diff
@@ -139,7 +139,7 @@
   "other-apps-by-developer": "Other apps by {{developer}}",
   "other-apps-by-projectgroup": "Other apps in the {{projectgroup}} group",
   "loading": "Loading…",
-  "could-not-find-match-for-search": "Could not find a match for search.",
+  "could-not-find-match-for-search": "Could not find a match for the search term.",
```
wjt, line 142: This is an improvement IMO but "search term" seems a bit like jargon (and what if there is more than one term?)

How about:

> Could not find any apps matching your search.

wjt (commented):
Just one suggestion :

razzeee:
This seems to need a rebase

kbdharun:
> This seems to need a rebase

Rebase and force push? (Wouldn't Squash & Merge work?)

PR #1724, opened 2023-07-31
Add developer invite system

jameswestman:
Adds a system for inviting other developers to a direct upload app.

There is a new Developers tab in the app management page for direct upload apps only. From there, the primary developer can invite or remove other developers, and non-primary developers can leave the app from there as well.

Each user account gets a random, unique invite code displayed on the Developer tab of their account page. To invite a developer, ask them for their code and paste it into the invite dialog. This way, all invites are associated with an existing user account, and we don't have to deal with the UX or backend complexities of creating an account and accepting an invite at the same time.

You must accept the Publisher Agreement to accept an invite.

Emails are sent:
- To the invited user when they receive an invite
- To all developers (including the new one) when an invite is accepted

There is a bug where developers can have a display_name of NULL, and they won't show up in the developers tab. This is probably best solved by overhauling some of the account management code and introducing a "main account" concept.


backend/app/logins.py
```diff
@@ -849,7 +850,10 @@ def continue_oauth_flow(
         # in then create a user
         user = login.user
         if user is None:
-            user = models.FlathubUser(display_name=provider_data["name"])
+            user = models.FlathubUser(
+                display_name=provider_data["name"],
+                invite_code=secrets.token_urlsafe(6),
```
razzeee, line 855: Wondering, if we could generate a word combination instead? `styled hatchet stoke blooms` for e.g. and I guess it could fail, when we insert here?

razzeee:
> There is a bug where developers can have a display_name of NULL, and they won't show up in the developers tab. This is probably best solved by overhauling some of the account management code and introducing a "main account" concept.

Can you create an issue for that?

I would like some backend tests, but I haven't had a look at how complex that part is

backend/app/models.py
```diff
@@ -570,6 +578,8 @@ class DirectUploadAppDeveloper(Base):
     )
     is_primary = Column(Boolean, nullable=False)
 
+    is_invite_pending = Column(Boolean, nullable=False, server_default=text("false"))
```
razzeee, line 581: I'm actually wondering, if we should manage that in a different (but mostly identical) table. As that would decrease the likelihood of giving wrong access when we change stuff up.

It also might allow us, to lock down the table constraints better.

frontend/pages/apps/manage/[appId]/accept-invite/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  params: { appId },
+}) => {
+  const app = await fetchAppstream(appId as string)
+
+  return {
+    props: {
+      ...(await serverSideTranslations(locale, ["common"])),
+      app: app ?? { id: appId, name: appId },
+    },
+    revalidate: 900,
```
Comment (razzeee) [inline frontend/pages/apps/manage/[appId]/accept-invite/index.tsx:110]: My gut feeling is, that this is wrong and likely not needed. We probably don't want to prebuild these pages

backend/app/config.py
```diff
@@ -20,6 +20,9 @@ class Settings(BaseSettings):
     datadir: str = os.path.join(ROOT_DIR, "data")
     stats_baseurl: str = "https://hub.flathub.org/stats"
     session_secret_key: str = "change-me-for-production"
+
+    frontend_url: str = "http://localhost:3000"
```
razzeee, line 24: One thing @barthalion needs to set, when we merge this

backend/app/emails.py
```diff
@@ -29,6 +29,8 @@
 
 class EmailCategory(str, Enum):
     SECURITY_LOGIN = "security_login"
+    DEVELOPER_INVITE = "developer_invite"
+    DEVELOPER_INVITE_ACCEPTED = "developer_invite_accepted"
```
razzeee, line 33: Makes me wonder, if we should send the inviting party an email, when the invitee declines.

backend/app/emails.py
```diff
@@ -130,6 +135,17 @@ def send_email(info: EmailInfo, db):
         for user in by_github_repo:
             messages.append(_create_message(user, info, db))
 
+        direct_upload_app = models.DirectUploadApp.by_app_id(db, info.app_id)
+        if direct_upload_app is not None:
+            by_direct_upload = (
+                db.session.query(models.DirectUploadAppDeveloper, models.FlathubUser)
```
razzeee, line 141: Wondering, if this should be on the model

frontend/public/locales/en/common.json
```diff
@@ -271,6 +271,24 @@
...
+  "decline": "Decline",
+  "invite-not-found": "Invite not found",
+  "invite-not-found-description": "The invite you are trying to accept or decline does not exist. It might have been cancelled or declined.",
+  "invited-apps": "Invitations",
+  "invite": "Invite",
+  "invite-code": "Your invite code is <1>{{inviteCode}}</1>. Give this code to another developer to allow them to invite you to their app.",
+  "invite-pending": "Invite pending",
+  "primary-developer": "Primary developer",
+  "invite-developer": "Invite developer",
+  "invite-developer-description": "Enter the developer's invite code to invite them to become a developer of this app. Their developer code is found on the Developers tab of their account page.",
```
razzeee, line 286:
```suggestion
  "invite-developer-description": "Enter the developers invite code to invite them to become a developer of this app. They can find the code on the Developers tab of their account page.",
```

razzeee: This also makes me wonder, if we should inject the translation for `Developers` here, to make sure it's not translated differently

frontend/pages/apps/manage/[appId]/accept-invite/index.tsx
Comment (jameswestman) [reply frontend/pages/apps/manage/[appId]/accept-invite/index.tsx:110]: What should I use instead? I copied this from the manage page.

Comment (razzeee) [reply frontend/pages/apps/manage/[appId]/accept-invite/index.tsx:110]: I think you should be able to drop the line completely and it should be fine.

backend/app/models.py
jameswestman: Good idea. It might also make it easier to implement ownership transfer.

frontend/public/locales/en/common.json
jameswestman: I think "developer's" is correct? Fixed the others.

razzeee: Reading up on this, I think you're right

backend/app/logins.py
```diff
@@ -939,11 +943,18 @@ def get_userinfo(login: LoginStatusDep):
...
+        user.invite_code = secrets.token_urlsafe(6)
+        db.session.commit()
+
     ret = {
         "is-moderator": user.is_moderator,
         "displayname": user.display_name,
         "dev-flatpaks": set(),
         "owned-flatpaks": set(),
+        "invited-flatpaks": set(),
+        "invite-code": user.invite_code,
```
razzeee, line 957: Would be nice to also add this to the type in `Login.ts`

backend/app/logins.py
jameswestman: Hmm, not sure if the UX is much better, especially for i18n? We should at least split the characters into chunks though.

razzeee: Yeah, that sounds like a neat compromise

backend/alembic/versions/c0472125ea51_invite_developers.py
```diff
@@ -0,0 +1,44 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('directuploadappinvite',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('app_id', sa.Integer(), nullable=False),
+    sa.Column('developer_id', sa.Integer(), nullable=False),
+    sa.Column('is_primary', sa.Boolean(), nullable=False),
```
razzeee, line 25: I found the inclusion of this field surprising. Is this really used/needed? I feel like that design might be something we would want to talk about. I at least can't come up with a sane one.

I think making somebody to be the primary should be a separate action and happen instantly, without needing the receiving part to accept.

razzeee (approved):
I think we're almost there, there's just one last thing that I found.

backend/alembic/versions/c0472125ea51_invite_developers.py
jameswestman: Yeah, I intended for that to be used to invite someone transfer primary ownership, but didn't implement it yet. It should probably be a separate action, but I'm not sure being able to dump maintainership on someone without their approval is a good idea.

razzeee: Let's remove the field and do it later?

frontend/pages/my-flathub.tsx
```diff
@@ -42,6 +44,25 @@ export default function Userpage({
           <div className="space-y-12">
             <UserApps variant="dev" />
 
+            {!IS_PRODUCTION && (
```
razzeee, line 47: I've started to guard cases like this with is production or is moderator for now.

razzeee:
I'll merge this tomorrow, if nobody speaks up.

PR #1725, opened 2023-07-31
backend: Use a model for the login state

jameswestman:
Use a class instead of a dictionary, so we get better syntax highlighting/completion. I implemented __getitem__ so the old dictionary style still works.


razzeee (approved):
As far as I can tell, this seems like a sensible improvement. Good work.

PR #1780, opened 2023-08-11
Add default account & clean up accounts logic

### LINKED ISSUE #1760: Improve connected accounts logic
The way Flathub stores data from connected accounts is a bit scattered and hard to navigate and update.

For example, we initialize a FlathubUser's display_name from the account they log in with first, and never update it. However, GitHub doesn't require a display name (you can have just a username) so sometimes this field is blank.

Also, we have 5 different login providers and could add more in the future, and it's getting unwieldy. We should add abstractions in the backend to make it easier to get a list of all a user's connected accounts, and have a common interface for them.

Finally, we should consider storing email addresses. We currently collect them again every time we need to send an email.

jameswestman:
Added some functions for unified handling of connected accounts. We can get a connected account by FlathubUser+provider, and we can get a list of all connected accounts.

Also added a "default_account" column to FlathubUser and a method to get the default account. It falls back to the first account in the list if no default has been set.

Finally, we now store display_name and email per account, and the email system uses the email address and display name from a person's default account when emailing them.

Fixes #1760.


backend/app/emails.py
```diff
@@ -91,8 +63,14 @@ def _create_html(info: EmailInfo, app_name: str, email: str, user_display_name:
...
     user: models.FlathubUser, info: EmailInfo, db
-) -> tuple[str, MIMEText]:
-    email = _get_email_address(user, db)
+) -> tuple[str, MIMEText] | None:
+    user_default_account = user.get_default_account(db)
+
+    email = user_default_account.email
+    if email is None:
+        # User doesn't have an email address on file
+        return None
```
razzeee, line 72: Wondering, if we need to log here

backend/app/models.py
```diff
@@ -37,6 +46,31 @@ class FlathubUser(Base):
...
     TABLES_FOR_DELETE = []
 
+    def connected_accounts(self, db) -> list["ConnectedAccount"]:
+        result = []
+        for table in ConnectedAccountTables:
+            if account := table.by_user(db, self):
+                result.append(account)
+        return result
+
+    def get_connected_account(self, db, provider: str) -> Optional["ConnectedAccount"]:
```
razzeee, line 56: Can we type the provider? Maybe with a union? Not sure if that works in python.

backend/app/logins.py
```diff
@@ -1106,6 +1076,24 @@ def do_agree_to_publisher_agreement(login: LoginStatusDep):
     db.session.commit()
 
 
+@router.post("/change-default-account", status_code=204)
+def do_change_default_account(
+    provider: str,
```
razzeee, line 1081: I guess this could also be a union, so that it shows in the docs

backend/app/logins.py
```diff
@@ -458,6 +458,15 @@ def start_oauth_flow(
     }
 
 
+@dataclass
+class ProviderInfo:
+    id: str
```
razzeee, line 463: Wondering, if we could get away with an int here, but it's probably risky

razzeee (approved):
Just a minor typing thing

backend/app/models.py
jameswestman: Changed it to use a string enum.

PR #1809, opened 2023-08-17
Parse summary of platforms

### LINKED ISSUE #239: Expose summary API for all Apps
I'm the developer of [flathub-data-viewer](https://jakobdev.github.io/flathub-data-viewer), a little tool which shows basic informations about Flathub. It can parse the Appstream Data directly, since it is just a XML file. But for things like Permissions and BaseApp it relays on the summary API of Flathub. But the API currently only supports Desktop Apps. Would be nice, if you could support all Types of Apps.

razzeee:
Closes #239

### DIFF
--- a/backend/app/summary.py
+++ b/backend/app/summary.py
@@ -1,13 +1,11 @@
-import configparser
 import json
-import re
 import struct
 import subprocess
 from collections import defaultdict
 
 import gi
 
-from . import config, db, search, utils
+from . import config, configParserCustom, db, search, utils
 
 gi.require_version("OSTree", "1.0")
 from gi.repository import Gio, GLib, OSTree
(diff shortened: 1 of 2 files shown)


backend/app/summary.py
```diff
@@ -272,14 +266,17 @@ def update():
...
         if "Build" in parser:
             build = parser["Build"]
             if "built-extensions" in build:
-                built_ext = build["built-extensions"].split(";")
-                for ext in built_ext:
+                if isinstance(build["built-extensions"], tuple):
+                    built_ext = ";".join(build["built-extensions"]).split(";")
+                else:
+                    built_ext = build["built-extensions"].split(";")
+                for ext in filter(len, built_ext):
```
razzeee, line 279: This is a slight fix for the reverse_lookup, as we could end up with empty keys here before this

PR #1820, opened 2023-08-20
Restore language after login

### LINKED ISSUE #1220: Language switches back to English after logging in
After having set a language other than English, and successfully logging in, the language switches back to English.

razzeee:
This sets a cookie on the providers page, containing the locale and restores it, after we have been redirected from the login provider.

Ideally we could send the locale to the login provider too, but that would  need changes to the redirect config, which I don't have access to.

Closes #1220


razzeee:
We probably want to keep an eye out after merging, for login regressions, as I have seen github asking to permit acceess multiple times during development. Hopefully that's just a side effect of development.

PR #1856, opened 2023-08-29
Fix some safety oversights

### LINKED ISSUE #1854: Safety rating inconsistent with GNOME Software 45
Since [Software !1712](https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/1712), free software apps with huge sandbox holes are only classified as "Potentially Unsafe" and only proprietary apps are "Unsafe".

razzeee:
Closes #1854

### DIFF
--- a/frontend/src/safety.ts
+++ b/frontend/src/safety.ts
@@ -122,13 +122,13 @@ export function getSafetyRating(
     )
   ) {
     appSafetyRating.push({
-      safetyRating: SafetyRating.unsafe,
+      safetyRating: SafetyRating.potentially_unsafe,
       description: "uses-session-services",
       showOnSummary: true,
     })
   }
 
-  // decives
+  // devices
   if (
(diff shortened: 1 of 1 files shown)


kra-mo:
Is the `Unsafe` rating ever used now? Glancing over the code, it seems like it isn't.

VSCode, which is both proprietary and unsandboxed is still only marked as `Potentially Unsafe`.

kra-mo:
image

razzeee:
> Is the `Unsafe` rating ever used now? Glancing over the code, it seems like it isn't.
> 
> VSCode, which is both proprietary and unsandboxed is still only marked as `Potentially Unsafe`.

It's not used. I think the only case where g-s uses it, would be apps that are closed source, when they are downrated. See https://gitlab.gnome.org/GNOME/gnome-software/-/blob/main/src/gs-safety-context-dialog.c?ref_type=heads#L396

I'm not so sure, I like that approach. But might just have been due to no elegant way to do it, with the way I've set it up.

PR #1857, opened 2023-08-29
Fix typo: `aquire` -> `acquire`

guihkx:


razzeee:
I think this should only touch the tsx and the english json file. Otherwise, we might get conflicts / missed translations otherwise.

guihkx:
I don't see how that would happen, but then again I'm not familiar with the codebase.

So, should I keep the typo in the name of the property (i.e. `can-aquire-arbitrary-permissions`)?

razzeee:
No, you can change it, but only in the common.json for `en`. Our translation system will then drop all non matching keys.

razzeee:
Thanks for working on this

PR #1863, opened 2023-08-29
Fix some license display edge cases

sgvictorino:
I found these via `org.chromium.Chromium` (see its [`project_license`](https://github.com/flathub/org.chromium.Chromium/blob/master/org.chromium.Chromium.metainfo.xml#L10)).


razzeee:
Yikes, that's a good find. I kinda want to add a test case for this. So I might push something to your branch.

razzeee:
Thanks for working on this

PR #1880, opened 2023-08-31
Refactor onChange events in search filters

liiiipe:


razzeee:
You'll need to run the formatter, before I can review this :)

PR #1906, opened 2023-09-06
Fix download stats from EOL Apps

JakobDev:
#1842 breaks getting the stats from EOL Apps. This should hopefully fix it, but I haven't tested it.


razzeee:
Thank you, seems to work fine. As long as we don't have non apps with EOL that we would want to show. But that's for when we need it.

PR #1936, opened 2023-09-10
fix: use sans-serif as fallback typeface

SethFalco:
As a user who browses the web with downloadable fonts disabled, Flathub looks shoddy because a fallback to a generic font family wasn't included.

As Inter is a sans-serif font, `sans-serif` should be appended to the end so the browser/OS knows to grab a sans-serif font. Due to different font metrics that may arise, I also removed the `max-width` that Tailwind added so that wider typefaces don't wrap to the next line.

## Before

image

## After

image

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -56,7 +56,7 @@ export default function Home({
       <NextSeo title={t("home")} description={t("flathub-description")} />
       <div className="max-w-11/12 mx-auto my-0 mt-12 w-11/12 space-y-10 2xl:w-[1400px] 2xl:max-w-[1400px]">
         <div className="flex justify-between gap-3">
-          <div className="prose dark:prose-invert">
+          <div className="prose dark:prose-invert max-w-none">
             <h1 className="mb-0 text-4xl font-extrabold">
               {t("the-linux-app-store")}
             </h1>
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -26,6 +26,7 @@ import { MotionConfig } from "framer-motion"
 
 const inter = Inter({
   subsets: ["latin"],
+  fallback: ["sans-serif"],
 })
 
 const App = ({ Component, pageProps }: AppProps) => {


frontend/pages/index.tsx
```diff
@@ -56,7 +56,7 @@ export default function Home({
       <NextSeo title={t("home")} description={t("flathub-description")} />
       <div className="max-w-11/12 mx-auto my-0 mt-12 w-11/12 space-y-10 2xl:w-[1400px] 2xl:max-w-[1400px]">
         <div className="flex justify-between gap-3">
-          <div className="prose dark:prose-invert">
+          <div className="prose dark:prose-invert max-w-none">
```
razzeee, line 59: I don't think we want this, there is a good reason, that the prose classes break up text after a while

razzeee (changes requested):

frontend/pages/index.tsx
SethFalco: Willing to remove it, but just want to confirm some details.

Generally speaking, you're right. The prose class ensures that everything is a comfortable width for reading. However, that doesn't really apply in this case because this is just a strapline, not paragraphs.

Meanwhile, Tailwinds' documentation notes that this is not always the desired behavior and circumventing it is supported behavior.

> Each size modifier comes with a baked in max-width designed to keep the content as readable as possible. This isn’t always what you want though, and sometimes you’ll want the content to just fill the width of its container.
> 
> — https://tailwindcss.com/docs/typography-plugin#overriding-max-width

From what I've tested, this does not impede or break other parts of the website. The `max-width` applied by prose actually introduces a problem rather than solving one.

If you still think it should be removed, I'll amend the commit, but I believe we should remove the max-width.

razzeee: Yeah, I think it's fine here. Looked at it on my phone earlier.

razzeee (approved):

razzeee:
If they don't have a `revalidate`, they get build on deploy. Pretty sure that would bust the cache. Not sure what could cause it.

razzeee:
Talked about this with barth, and it should be fine to switch those pages back to `revalidate`

PR #1937, opened 2023-09-10
docs: migrate to docker compose v2

SethFalco:
* Replaces references to Docker Compose v1 to Docker Compose v2.
* Fixes a few typos in the README.

Docker Compose v2 has been out for some time, and Docker Compose v1 (`docker-compose`) no longer receives updates.

> From July 2023 Compose V1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.
> 
> — https://docs.docker.com/compose/ 

I think it makes sense to push the documentation and commands over to the v2 equivalents now, since this is what most engineers should have installed now.

## Related

* Similar to https://github.com/nextcloud/cookbook/pull/1772
* Similar to https://github.com/juliushaertl/nextcloud-docker-dev/pull/209

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Pull requests are welcome. Please, create an issue first explaining what you wan
 
 ### Development
 
-You'll need `git`, `yarn`, `docker` and `docker-compose`.
+You'll need `git`, `yarn`, `docker` and `docker-compose-plugin`.
 
 Go to the folder where you manage your projects and checkout this project.
 
@@ -22,7 +22,7 @@ Then start the backend:
 
 ```sh
 cd backend
-docker-compose up
+docker compose up
 ```
 
 You might need to use `--build` to rebuild the images, when dependencies change.
--- a/backend/README.md
+++ b/backend/README.md
@@ -1,7 +1,7 @@
 # Flathub website backend
 
 This is the fastapi based backend for https://www.flathub.org
-Not the backend for flathub itself. Go here, if your looking for that https://github.com/flathub/flathub
+Not the backend for flathub itself. Go here, if you're looking for that https://github.com/flathub/flathub
 
 ## Development
 
@@ -15,7 +15,7 @@ Not the backend for flathub itself. Go here, if your looking for that https://gi
 Start the database:
 
 ```bash
-docker-compose up
+docker compose up
 ```
 
 You need to seed the database:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -14,15 +14,15 @@ Pull requests are welcome. Please, create an issue first explaining what you wan
 
 ### Development
 
-You'll need `git`, `yarn`, `docker` and `docker-compose`.
+You'll need `git`, `yarn`, `docker` and the Docker Compose plugin.
```
razzeee, line 17: I would actually drop the name `docker-compose-plugin` here. That's what it's called for fedora anyway. Had to find it in the docs, as I could never make `docker compose` without a dash work.

https://docs.docker.com/engine/install/fedora/#set-up-the-repository

razzeee:
You seem to have missed the mentions of docker-compose in the `tasks.json`

README.md
SethFalco: Hmm, I'm a bit confused tbh. :thinking: 

> That's what it's called for fedora anyway.

In most Linux distros, the package is called `docker-compose-plugin` I believe.

<details>
<summary><code>$ apt list docker-compose-plugin</code></summary>

```
Listing... Done
docker-compose-plugin/bookworm,now 2.21.0-1~debian.12~bookworm amd64 [installed]
N: There are 7 additional versions. Please use the '-a' switch to see them.
```

</details>

> as I could never make `docker compose` without a dash work.

Assuming you followed the docs you shared in the past few months, I believe that would install Docker Compose v2. So, a plugin for Docker rather than a separate utility.

If you wouldn't mind, would you be willing to share the output of:

```sh
docker --version
docker-compose --version
docker compose version
```

razzeee: Sorry, I think I wasn't very clear. It works fine now.

I did have to dig through the docs to get it installed earlier. Before that I wasn't aware, that `docker-compose` is v1 and `docker compose` is v2 and you need to install something separate. So every time anyone came up with commands without a dash, it was very confusing why it was working for them but not for me.

After running ` sudo dnf install docker-compose-plugin` earlier today it just started working. So I think we should name that package in the readme, to give a hint. As just saying `Docker Compose plugin` might send most people to a search engine.  Saying `docker-compose-plugin` might tip of some of the more experienced users.

SethFalco: Ahh, this all makes sense, thanks for clarifying!
I can relate actually, I had the same experience for a few months too. ^-^'

Making the change now.

razzeee:
Thanks for looking into this.

Issue #1818, opened 2023-08-20
App content moderation


razzeee:
The plan here is to have a panel only shown to:

1. App authors (readable)
2. App content moderators (changeable)

This panel should link to https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines as much as possible.

1. App authors should be able to see how their app has been reviewd and be able to look at the docs to potentially improve

2. Content moderators should be able to change the status of a moderation item (keep it simple) 

![image](https://github.com/flathub/website/assets/5943908/ca68cf78-daa6-4e24-9c87-5f5ccc191ad8)

This data should in the first instance only be surfaced in the logged in state for app authors. We might later use it, to inform the landing page or use it to choose apps for special listings.

razzeee:
This has been done

PR #1977, opened 2023-09-16
Implement quality guideline ui for quality moderators

razzeee:
Start of #1818


razzeee:
This is only the moderator facing side of things

grafik

grafik

backend/app/worker.py
```diff
@@ -190,3 +190,32 @@ def send_email(email):
...
+
+        for appid in appids_for_frontend:
+            # Check app name lenght
+            if value := db.get_json_key(f"apps:{appid}"):
+                models.QualityModeration.upsert(
+                    sqldb, appid, 8, len(value["name"]) <= 20, None
+                )
+
+            # Check app summary lenght
+            if value := db.get_json_key(f"apps:{appid}"):
```
jameswestman, line 218: Could combine these two if blocks

backend/app/worker.py
```diff
@@ -190,3 +190,32 @@ def send_email(email):
...
+        ]
+
+        if not appids_for_frontend:
+            return
+
+        for appid in appids_for_frontend:
+            # Check app name lenght
+            if value := db.get_json_key(f"apps:{appid}"):
+                models.QualityModeration.upsert(
+                    sqldb, appid, 8, len(value["name"]) <= 20, None
```
jameswestman, line 214: I'm not a huge fan of using "magic numbers" for the guideline IDs. Strings would probably be clearer.

Another thing to consider is it might be easier to store the guidelines in code within the backend (or even the frontend), rather than in the database.

backend/app/models.py
```diff
@@ -1317,3 +1320,128 @@ class ModerationRequest(Base):
...
+                        guideline_id=guideline.id,
+                        app_id=appid,
+                        updated_at=None,
+                        updated_by=None,
+                        passed=None,
+                        comment=None,
+                    )
+                )
+
+        # Add guideline data to quality moderation
```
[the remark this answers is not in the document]
razzeee: Good catch, I think I moved that around at some point, so you're correct

backend/app/worker.py
razzeee: Should be better now

backend/app/worker.py
razzeee: I think I had a reason for this, but I can't remember.

I'm also not sure, if moving stuff into code (backend) would improve. I guess it's easier, as we have one deployed system, but this code is ready and this makes it arguably easier to extend.

Issue #1979, opened 2023-09-17
Re-assessing application safety levels


mihnea-radulescu:
### Issue

Application safety due to requested permissions is currently assessed by looking into the permissions in **isolation**.

**Example**: an application requesting the _unsafe_ home folder read permission, without requesting other _unsafe_ permissions, is granted a _Potentially unsafe_ rating, even though read permissions have no potential of compromising user data, in the absence of write or send permissions.

### Fix

Application safety due to requested permissions should be assessed by looking into the permissions in **conjunction**, rather than in isolation.

**Example**: an application requesting the _unsafe_ home folder read permission, without requesting other _unsafe_ permissions, should be granted a _Probably safe_ rating.

### PR which fixes the issue

[PR #1980](https://github.com/flathub/website/pull/1980)

mihnea-radulescu:
Closed due to a lack of interest from the GNOME community.

PR #1980, opened 2023-09-17
Feature - Data containment levels

mihnea-radulescu:
Fixes the issue [#1979](https://github.com/flathub/website/issues/1979).

### DIFF
--- a/frontend/src/safety.ts
+++ b/frontend/src/safety.ts
@@ -607,7 +607,8 @@ function specificFileHandling(
     appSafetyRating.push(...specificFileHandlingAppSafetyRating)
 
     const highestDataContainmentLevel = Math.max(
-      ...Object.values(specificFileHandlingAppSafetyRating).map((x) => x.dataContainmentLevel),
+      ...Object.values(specificFileHandlingAppSafetyRating).map(
+        (x) => x.dataContainmentLevel),
     )
 
     appSafetyRating.push({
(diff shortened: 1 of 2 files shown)


frontend/src/safety.ts
```diff
@@ -550,11 +591,19 @@ function specificFileHandling(
       )
       if (partialMatch.length > 0 && fileSystem.partialMatchKey) {
         partialMatch.forEach((x) => {
+          const translationResult = readWriteTranslationKey(x)
+
+          highestDataContainmentLevel = Math.max(
```
razzeee, line 596: Might be nicer, to just do this at the end before line 617 based on the contents of `appSafetyRating`?

razzeee:
I like this as an idea, but I would like to talk with the gnome software maintainers about this, would you be open to creating an issue on their bug tracker and reaching out?

frontend/src/safety.ts
mihnea-radulescu: Thanks for the suggestion, implemented!

mihnea-radulescu:
Thank you for the feedback, @razzeee!

I have created [an issue](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2340) on Gnome GitLab.

mihnea-radulescu:
Closed due to a lack of interest from the GNOME community.

PR #1981, opened 2023-09-17
Use react-query in saved card components

kymckay:
Recently started using react-query in a personal project and thought I'd come back and clean up my old mess.

Edit: Found a few broken endpoints while testing this too, but moved those fixes to: #1982

### DIFF
--- a/frontend/src/components/payment/cards/SavedCards.tsx
+++ b/frontend/src/components/payment/cards/SavedCards.tsx
@@ -1,54 +1,43 @@
 import { useTranslation } from "next-i18next"
-import {
-  FunctionComponent,
-  ReactElement,
-  useCallback,
-  useEffect,
-  useState,
-} from "react"
+import { FunctionComponent, ReactNode, useCallback, useState } from "react"
 import { getPaymentCards } from "../../../asyncs/payment"
-import { useAsync } from "../../../hooks/useAsync"
 import { PaymentCard } from "../../../types/Payment"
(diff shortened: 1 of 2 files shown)


frontend/src/components/payment/cards/SavedCards.tsx
```diff
@@ -1,54 +1,43 @@
...
 import CardInfo from "./CardInfo"
 import DeleteCardButton from "./DeleteCardButton"
+import { useQuery } from "@tanstack/react-query"
 
 const SavedCards: FunctionComponent = () => {
   const { t } = useTranslation()
 
-  const [cards, setCards] = useState<PaymentCard[]>()
+  const walletQuery = useQuery<PaymentCard[]>({
+    queryKey: ["/walletinfo"],
```
kymckay, line 14: One thought I just had: this probably needs to also include the logged in user ID to avoid using cached results after switching user (edge case, but possible)

razzeee: I guess we might just want to drop the whole cache, when a user logs out.
https://tanstack.com/query/v4/docs/react/reference/QueryClient#queryclientclear or maybe
 https://tanstack.com/query/v4/docs/react/reference/QueryCache#querycacheclear

Not sure if there might be a better way

razzeee: I'll merge this for now, as it looks trivial enough and will create an issue to tackle the cache problem

PR #1997, opened 2023-09-19
Refresh needs the user to have a github account connected

razzeee:
This bails out, if the user doesn't have a github account

### DIFF
--- a/backend/app/logins.py
+++ b/backend/app/logins.py
@@ -994,11 +994,16 @@ def get_userinfo(login: LoginStatusDep):
 
 @router.post("/refresh-dev-flatpaks")
 def do_refresh_dev_flatpaks(request: Request, login: LoginStatusDep):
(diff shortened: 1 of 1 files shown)


backend/app/logins.py
```diff
@@ -994,12 +994,17 @@ def get_userinfo(login: LoginStatusDep):
 
 @router.post("/refresh-dev-flatpaks")
 def do_refresh_dev_flatpaks(request: Request, login: LoginStatusDep):
-    if login.state == LoginState.LOGGED_OUT:
+    if login.state == LoginState.LOGGED_OUT or login.user is None:
         return Response(status_code=401)
```
jameswestman, line 998: Might do `raise HTTPException(status_code=401, detail="not_logged_in")` to provide a little more detail. Same below.

razzeee: Updated

PR #1999, opened 2023-09-20
feat: migrate to stripe API version 2023-08-16

kymckay:
Found some time to look into this, not much has changed for our integration. Just as follows.

- PaymentIntent no longer has `charges` property, replaced with `latest_charge`.
- Need to expand `latest_charge` when retrieving PaymentIntent to get receipt URL.
- Updated the Stripe setup docs to cover correctly seeding account for transfer testing.

I tested creating an express account, configuring an app for payment, purchasing the app and then processing the pending transfers. All are working.

closes #1972


razzeee:
Thank you

PR #2003, opened 2023-09-22
Implement quality moderation on the moderator side of things

razzeee:


backend/app/models.py
```diff
@@ -1317,3 +1320,66 @@ class ModerationRequest(Base):
     handled_at = mapped_column(DateTime)
     is_approved = mapped_column(Boolean)
     comment = mapped_column(String)
+
+
+class QualityModeration(Base):
+    """A moderation request for a quality guideline"""
```
razzeee, line 1326: This needs to be updated, as it's not a request

PR #2024, opened 2023-09-27
Drop query client cache on logout

### LINKED ISSUE #1984: Drop query cache on logout
One thought I just had: this probably needs to also include the logged in user ID to avoid using cached results after switching user (edge case, but possible)

_Originally posted by @kymckay in https://github.com/flathub/website/pull/1981#discussion_r1328157905_

razzeee:
We might have cached user specific calls, let's drop them on logout

Closes #1984


kymckay:
Just to check: Does this imply the query client is a singleton?

Otherwise I would've thought this has no effect since it's not the same instance as the hook gets from the `QueryClientProvider`.

razzeee:
not necessarily, if the cache is the same object/storage for every client/user (which it hopefully is to make use of it) it would be enough to drop that.

PR #2029, opened 2023-09-28
Copy: login → log in

cassidyjames:
Similar to #2023, "login" is a noun while "log in" as the verb form. Since we don't appear to use the `login` string as a noun anywhere, we should explicitly use the verb form.


frontend/public/locales/en/common.json
```diff
@@ -74,7 +74,7 @@
   "an-error-occurred-server": "{{errorCode}} - An error occurred on the server.",
   "whoops": "Whoops!",
   "could-not-find-page": "We couldn't find the page you were looking for.",
-  "login": "Login",
+  "login": "Log In",
```
razzeee, line 77:
```suggestion
  "login": "Log in",
```
Not a fan of how it looks with the other casing

cassidyjames: It's used in titles, which are _typically_ in title case, but we should probably have a rule for that since we've not been consistent.

razzeee: For what it's worth 
image

(then again facebook uses the other one)

cassidyjames: Hm, yeah maybe a language thing, in which case it should be localized in the appropriate language files of course. 😄 For what it's worth, both the [GNOME HIG] and [KDE HIG] dictate header/title capitalization for this—for both the button and page title.

[GNOME HIG]: https://developer.gnome.org/hig/guidelines/writing-style.html#header-capitalization
[KDE HIG]: https://develop.kde.org/hig/style/writing/capitalization/#title-capitalization

PR #2084, opened 2023-10-09
feat: add codegen for frontend API client

### LINKED ISSUE #2083: Use codegen for frontend API consumption
We can use code generation to produce correct types and API functions by consuming the OpenAPI specification FastAPI exposes. This means the backend Pydantic models become the single source of truth for request and response data typing, which brings a few benefits:
- Increased chance of catching errors during development caused by bad/missing data and typing
- Reduced maintenance burden when changing or adding endpoints used by the frontend
- More consistency in the frontend type definitions and API usage functions
- Convenience for frontend development, no longer need to manage these types manually

Can migrate use of the old types over case-by-case and eventually remove them instead of a big-bang approach. I recently implemented this for a personal project using [openapi-typescript-codegen](https://github.com/ferdikoomen/openapi-typescript-codegen) and would be happy to create a PR here if you think this is desirable @razzeee ?

kymckay:
This is just a draft PR to demonstrate a simple use of the codegen. Here I'm demonstrating the use for the `GET /auth/login` endpoint.

Nice things:
- You can see that codegen outputs client services based on the tags given to each path operation, probably a good habit to get into anyway for the OpenAPI spec.
- By adding proper type information to the path operation function, the frontend gets perfect typing for free from the codegen - this has actually already resolved an unspecified type in one of these example uses. It also encourages strong typing on the API, another good habit to get into.
- By using codegen for the frontend client, we no longer need the manually defined fetcher for this endpoint.
- The codegen can output its clients backed by axios requests which we're already using, I've set the package.json script and new VSCode task to do so.

Ugly things:
- You can see that I'm setting the `OpenAPI.BASE` to be our API's base URI in `_app.tsx`. This feels kind of hacky to me, but works and seems to be the recommended approach.
- The function names for each path operation are currently a little verbose (e.g. `getLoginKindsAuthLoginGet`). It's because they're taken from the operation IDs in the OpenAPI spec. We can change how those are generated in the backend quite easily (to be only the function name for example), but they need to be unique across all paths so there's a trade off there of shifting responsibility on to the developer.

Would close #2083

### DIFF
--- a/backend/app/logins.py
+++ b/backend/app/logins.py
@@ -231,8 +231,13 @@ def login_state(request: Request) -> LoginInformation:
 LoginStatusDep = Annotated[LoginInformation, Depends(login_state)]
 
 
-@router.get("/login")
-def get_login_kinds():
+class LoginMethod(BaseModel):
+    method: str
+    name: str
+
+
+@router.get("/login", tags=["login"])
+def get_login_kinds() -> list[LoginMethod]:
     """
     Retrieve the login methods available from the backend.
 
@@ -244,22 +249,10 @@ def get_login_kinds():
     frontends with localisation may choose to render other text instead.
     """
     return [
-        {
-            "method": "github",
-            "name": "GitHub",
-        },
-        {
-            "method": "gitlab",
-            "name": "GitLab",
-        },
-        {
-            "method": "gnome",
-            "name": "GNOME GitLab",
-        },
-        {
-            "method": "kde",
-            "name": "KDE GitLab",
-        },
+        LoginMethod(method="github", name="GitHub"),
+        LoginMethod(method="gitlab", name="GitLab"),
+        LoginMethod(method="gnome", name="GNOME GitLab"),
+        LoginMethod(method="kde", name="KDE GitLab"),
     ]
 
 
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,7 +13,8 @@
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "postbuild": "next-sitemap",
-    "update-setup-instructions": "./setup-instructions.sh"
+    "update-setup-instructions": "./setup-instructions.sh",
+    "openapi-codegen": "openapi -i http://localhost:8000/openapi.json -o ./src/api -c axios"
   },
   "dependencies": {
     "@collapsed/react": "^4.0.1",
@@ -74,6 +75,7 @@
     "eslint-config-next": "^13.5.3",
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-prettier": "^5.0.0",
+    "openapi-typescript-codegen": "^0.25.0",
     "postcss": "^8.4.31",
     "prettier": "^3.0.0",
     "prettier-plugin-tailwindcss": "^0.5.4",
--- a/frontend/src/fetchers.ts
+++ b/frontend/src/fetchers.ts
@@ -245,10 +245,6 @@ export async function fetchSearchQuery(
   )
 }
 
-export async function fetchLoginProviders() {
-  return axios.get<LoginProvider[]>(LOGIN_PROVIDERS_URL)
-}
-
 export async function fetchGithubRequestOrgAccessLink() {
   return axios.get<{ link: string }>(REQUEST_ORG_ACCESS_LINK_GITHUB)
 }
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -86,5 +86,14 @@
                 "cwd": "${workspaceFolder}/backend"
             },
         },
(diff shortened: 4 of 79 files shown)


kymckay:
@razzeee Curious to hear your thoughts. I like that this encourages good habits and enforces strong typing. I don't love the current function naming output, but if we're okay with taking responsibility for defining unique operation IDs that could be improved.

From the failing CI, it looks like we would also need to target at least ES6 in tsconfig. Do we have any strong feelings about targeting newer versions of JS in our frontend builds?

kymckay:
- [ ] Check that this works as expected on an endpoint that uses authentication.

frontend/package.json
```diff
@@ -13,7 +13,8 @@
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "postbuild": "next-sitemap",
-    "update-setup-instructions": "./setup-instructions.sh"
+    "update-setup-instructions": "./setup-instructions.sh",
+    "codegen": "openapi -i http://localhost:8000/openapi.json -o ./src/api -c axios"
```
razzeee, line 17: I think I would like something alnog the lines of `openapi-codegen` more

razzeee:
> From the failing CI, it looks like we would also need to target at least ES6 in tsconfig. Do we have any strong feelings about targeting newer versions of JS in our frontend builds?

No, as long as it not causes any other problems.

frontend/src/fetchers.ts
```diff
@@ -245,10 +245,6 @@ export async function fetchSearchQuery(
   )
 }
 
-export async function fetchLoginProviders() {
-  return axios.get<LoginProvider[]>(LOGIN_PROVIDERS_URL)
```
razzeee, line 249: Removing `LOGIN_PROVIDERS_URL` seems to be missing?

backend/app/logins.py
```diff
@@ -231,8 +231,13 @@ def login_state(request: Request) -> LoginInformation:
...
 
 
-@router.get("/login")
-def get_login_kinds():
+class LoginMethod(BaseModel):
+    method: str
+    name: str
+
+
+@router.get("/login", tags=["login"])
```
razzeee, line 239: So without a tag, it just ends up in `DefaultService.ts`?

razzeee:
Seems like an improvement, especially cause I never was a fan of the fetcher or env files (for endpoints) anyway.

But I agree, the naming is harsh.

Still it's no graphql, but that's not in the cards anyway :/

backend/app/logins.py
kymckay: That's correct

frontend/src/fetchers.ts
kymckay: It's still currently used in the rest of the auth flow.

kymckay:
For the naming, in the backend FastAPI allows us to control how the operation IDs are generated:

```python
app = FastAPI(
    generate_unique_id_function=lambda x: x.name,
)
```

This snippet would result in just  `LoginService.getLoginKinds`.  Can also [manually set them per path](https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#openapi-operationid) if needed. So depends whether we value the guaranteed uniqueness over the more readable frontend calls.

There's also [OpenAPI Generator](https://openapi-generator.tech/) which might provide a better structured output. I'll give it a try for comparison too.

razzeee:
I'm a bit unsure about all those pythonic snake case names, but meh.

Worst part was the body I think

kymckay:
From a little more investigation, it does look like there are some shortcomings to this library: [1](https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1629), [2](https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1468). I reckon it's still better than the manual status quo - we could even consume a fork with some of the relevant features/fixes merged at a future point if we really need to.

I had a quick go at using OpenAPI Generator, but it's failing because some of the dependencies are using CommonJS modules and not ESM.

Will check the pain points of using an endpoint that requires authentication in a little bit and then can make a decision whether to merge.

kymckay:
Having experimented a little more with this, I'm no longer as convinced we should adopt it.

- To handle endpoints with authentication we need to do the same global override trick in `_app.tsx` which results in all requests including the session cookie, etc. Not the end of the world, but not ideal.
- I'm also having issues with react-query stuck in fetching state even though I can see the network request succeeding and returning data. Having triple checked the logic and trying with different endpoints, I think it may be due to the underlying `CancelablePromise` object the library creates (though I've not encountered the same behaviour where I'm using this in another project). Apparently cancellable promises are a construct from a scrapped JS spec proposal.
- The generated client is causing unhandled runtime errors when testing network failures instead of react-query catching the error into query state.

PR #2116, opened 2023-10-21
Implement app of the day automatic picking

razzeee:
This automatically picks an app of the day

### DIFF
--- a/backend/app/update.py
+++ b/backend/app/update.py
@@ -14,6 +14,7 @@ def register_to_app(app: FastAPI):
 async def update():
     worker.update.send()
     worker.update_quality_moderation.send()
+    worker.update_app_picks.send()
 
 
 @router.post("/stats", tags=["update"])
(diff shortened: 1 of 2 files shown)


razzeee:
Should be almost ready.

1. Auth in backend is broken
2. We might want to land the backend separately first
3. Needs a review

razzeee:
I will cut out the not user facing pieces and land them before the rest get's landed

backend/alembic/versions/f8b57ec1337e_add_position_to_apps_of_the_week.py
```diff
@@ -18,11 +18,13 @@
 
 def upgrade():
     # ### commands auto generated by Alembic - please adjust! ###
-    op.add_column('appsoftheweek', sa.Column('position', sa.Integer(), nullable=False))
+    op.add_column(
+        'apps_of_the_week', sa.Column('position', sa.Integer(), nullable=False)
```
razzeee, line 22: These changes are problematic and need to be ripped out again

PR #2117, opened 2023-10-22
feat: add codegen for frontend API client

### LINKED ISSUE #2083: Use codegen for frontend API consumption

kymckay:
Adds codegen using OpenAPI generator.

- While this only changes the simple login providers endpoint, I did test it with `/walletinfo` too which requires session cookie authentication. The codegen is suitable because it allows passing the axios request config via the function call, so we can include the session cookie on a per-request basis.
  - The reason I didn't include further changes here is to keep the PR scope tight on only introducing codegen. Changing `/walletinfo` quickly became a lot of required changes due to the knock on effect.
- I've included a shell script to perform codegen and tidy up. OpenAPI generator seems to output in the shape of an npm package with additional artifacts we don't want.
- I decided to introduce this central `api.ts` module to instantiate the API objects in one place instead of littering the code with new instantiations every time we want to make a request.

closes #2083


kymckay:
This supersedes #2084.

Note that this library also uses the endpoint tags to generate separate client service objects. In my opinion, it's a good practice for us to be adding tags anyway.

backend/app/logins.py
```diff
@@ -231,8 +231,13 @@ def login_state(request: Request) -> LoginInformation:
 LoginStatusDep = Annotated[LoginInformation, Depends(login_state)]
 
 
-@router.get("/login")
-def get_login_kinds():
+class LoginMethod(BaseModel):
+    method: str
```
razzeee, line 235: This feels like it should be an enum/union, but probably doesn't matter

frontend/openapi-codegen.sh
```diff
@@ -0,0 +1,15 @@
+# Clean slate
+rm -rf ./src/codegen
+
+npx @openapitools/openapi-generator-cli generate \
+  -i http://localhost:8000/openapi.json -o src/codegen \
+  -g typescript-axios --additional-properties=supportsES6=true
+
+# Tidy up unnecessary artifacts
+# Not interested in publishing as a separate npm package
```
razzeee, line 9: Wondering if there might be settings for this.

In general, I'm a bit split, if we even want to check this in. We could also just gitignore it and run it as a pre-start script. But it's not as explicit when something changes that way.

frontend/pages/my-flathub.tsx
```diff
@@ -126,12 +126,12 @@ export default function Userpage({
 
 // Need available login providers to show options on page
 export const getStaticProps: GetStaticProps = async ({ locale }) => {
-  const { data: providers } = await fetchLoginProviders()
+  const providers = await loginApi.getLoginMethodsAuthLoginGet()
```
razzeee, line 129: not a fan of the names it generates, but I guess, we could also tackle this later on. still might be easier to do now.

razzeee:
Can we somehow ignore paths? We don't need the `compat` endpoints for example

razzeee:
Needs a readme adjustment and maybe some hints, that codgen needs the backend to run?

backend/app/logins.py
```diff
@@ -231,8 +231,13 @@ def login_state(request: Request) -> LoginInformation:
...
+    name: str
+
+
+@router.get("/login", tags=["login"])
```
razzeee, line 239: Looking at the api structure this should be `auth`

```suggestion
@router.get("/login", tags=["auth"])
```

razzeee:
Would be cool, to format the files via something like this https://openapi-generator.tech/docs/file-post-processing#example

frontend/openapi-codegen.sh
kymckay: I thought there may be a setting also, but doesn't seem so in [the documentation](https://openapi-generator.tech/docs/generators/typescript-axios).

frontend/pages/my-flathub.tsx
kymckay: Similar to the other library, this name is based on the operation name in the OpenAPI spec. It's easy to change the formula for those in the backend, but for a valid spec they should each be unique which is why FastAPI is concatenating the function name + path + HTTP method.

Another option is the method [described in the FastAPI docs](https://fastapi.tiangolo.com/advanced/generate-clients/#preprocess-the-openapi-specification-for-the-client-generator) of preprocessing the spec before feeding it to the codegen.

frontend/openapi-codegen.sh
kymckay: Actually, I think there is a global property for this rather than generator specific. Will try that out.

kymckay: I've slightly improved this using the ignore file in the output directory. Less file churn that way too. We could .gitignore the remaining artifacts that we don't care for which are un-skippable.

frontend/pages/my-flathub.tsx
razzeee: I would have thought path + http method is enough to generate unique names

I guess we'll just need to rework the function naming in the backend then, probably after merging this. Stopping to include the get/post/whatever in the function name.

frontend/openapi-codegen.sh
razzeee: I'm not sure which ones you mean, I can't see anything in that folder that seems useless now

kymckay: The remaining files this script is still removing

backend/app/logins.py
razzeee: Yeah, I don't care enough about it / hate python too much

frontend/openapi-codegen.sh
razzeee: I think we should probably keep those around, they seem like cache/version info, so might be relevant on updates?

razzeee:
I would be happy like this. But I would squash it, as I've messed up a commit :)

The only thing I can think of now, is that the generator throws some error, we would likely want to look at.

kymckay:
I also think I've figured out how to ignore certain paths, will push in a moment

kymckay:
Pushed new generation structure. Annoyingly while it allows ignoring the compat api, it will still be preset as an export in index.ts which doesn't work. I still think this source structure is better for diffs.

There is an `apis` property to specify which ones to include by name instead, but that would require keeping up to date with the source. Feels like a use case oversight in this library.

kymckay:
One downside to this structure: https://github.com/OpenAPITools/openapi-generator/issues/15669

When types are no longer relevant the file will stick around. Feel free to revert that last commit.

PR #2119, opened 2023-10-22
Rework wallet info code

kymckay:
- Use new codegen in frontend for this endpoint
- Provide correct typing and respect it in backend
- Refactor checkout component to use react-query to fetch walletinfo

Notes:
- Also includes vending API codegen addition. This PR has highlighted to me that our codegen will have different results depending on if the backend is running with stripe environment set up or not. The remedy to that would be to align the fake wallet with the stripe one so the endpoints and return types are consistent.
- I should probably refactor and cleanup the checkout component state and flow. This was a pain to rework because of how tightly coupled everything there is.


frontend/src/components/payment/cards/CardInfo.tsx
```diff
@@ -3,11 +3,11 @@ import { useTheme } from "next-themes"
 import { FunctionComponent, MouseEventHandler } from "react"
 import ReactCountryFlag from "react-country-flag"
 import { IMAGE_BASE_URL } from "../../../env"
-import { PaymentCard } from "../../../types/Payment"
 import Image from "next/image"
+import { CardInfo as TCardInfo } from "src/codegen"
```
kymckay, line 7: I should probably rename this type in the backend so it doesn't conflict with this frontend component name.

I think `WalletCard` is more descriptive anyway.

razzeee:
Thanks, seems straigtforward and uncritical enough, will merge now.

PR #2133, opened 2023-10-23
Optional string type

razzeee:


frontend/src/codegen/model/current-values.ts
```diff
@@ -0,0 +1,20 @@
...
+ * https://openapi-generator.tech
+ * Do not edit the class manually.
+ */
+
+/**
+ *
+ * @export
+ * @interface CurrentValues
+ */
+export interface CurrentValues {}
```
razzeee, line 20: @kymckay any idea why this would generate wrong like this? I would probably just blame python :/

frontend/src/codegen/model/keys.ts
```diff
@@ -0,0 +1,20 @@
...
+ * https://openapi-generator.tech
+ * Do not edit the class manually.
+ */
+
+/**
+ *
+ * @export
+ * @interface Keys
+ */
+export interface Keys {}
```
razzeee, line 20: Same here

frontend/src/components/moderation/AppstreamChangesRow.tsx
```diff
@@ -41,9 +41,9 @@ const AppstreamChangesRow: FunctionComponent<Props> = ({ request }) => {
             <tr key={key}>
               <td>{key}</td>
               {!request.is_new_submission && (
-                <td>{request.request_data.current_values[key]}</td>
+                <td>{request.request_data.current_values[key] as string}</td>
```
razzeee, line 44: I'm forcing it to be a string then again, so that the frontend doesn't fail

frontend/src/codegen/model/current-values.ts
kymckay: @razzeee 

Few possible causes:
- The backend type can't be expressed in openapi spec (seems unlikely)
- FastAPI fails to express the backend type in the openapi spec
- OpenAPI Generator fails to support the way this type is expressed in the openapi spec

I'd check how the type appears in `openapi.json`, if that looks legit then it's probably the latter of those options - which seems to be the case here:

```json
"RequestData": {
  "properties": {
    "keys": {
      "additionalProperties": {
        "anyOf": [
          {
            "type": "string"
          },
          {
            "type": "null"
          }
        ]
      },
      "type": "object",
      "title": "Keys"
    },
    "current_values": {
      "additionalProperties": {
        "anyOf": [
          {
            "type": "string"
          },
          {
            "type": "null"
          }
        ]
      },
      "type": "object",
      "title": "Current Values"
    }
  },
  "type": "object",
  "required": [
    "keys",
    "current_values"
  ],
  "title": "RequestData"
},
```

razzeee: Docs on https://swagger.io/docs/specification/data-models/data-types/ (search for `null`) suggest, that it's generated wrong? Then again, it might not be valid in that case

razzeee: Then again, seems like that would also be buggy in the generator https://github.com/OpenAPITools/openapi-generator/issues/13445

kymckay: Good find, so FastAPI seems the likely culprit. Could test grabbing the json and editing this to use `nullable` for the string type to see if codegen then works.

razzeee: Doesn't seem to affect, what's generated at all, as far as I could test it

PR #2141, opened 2023-10-24
fix(size): replaced SI prefixes with IEC prefixes

### LINKED ISSUE #2139: Replace SI prefixes with IEC for units of information (ISO/IEC 80000)
This is the copy of the https://github.com/flatpak/flatpak/issues/5504, since Flatpak CLI and Flathub web both have the same issue.

To be more transparent when it comes to digital size of something (binaries/flatpak packages), flatpak (both CLI and web) should move from SI prefixes to IEC (binary) prefixes, in other words add a small `i` like so: `MB` → `MiB` (same with `G`, `K` etc.). Right now, just looking at the 69 MB, user can't know for sure if it is `69 Mebi × 1024 Kibi × 1024 Byte × 8 bit = 578813952 bit` or `69 Mega × 1000 Kilo × 1000 Byte × 8 bit = 552000000 bit`, which already an almost 5% difference in size.

Many CLI/GUI programs already use binary prefixes (I'm so happy), but there are still many others that don't (`:(`).

AFAIK, this also intentionally (ab)used by marketing teams to sell something with 2 TB of storage, but it's definitely not the 2 TiB that everyone wants to think. Why do I think this way? Because my peers and teachers/professors don't even know that there is a thing called "Kibibyte" (and we are talking about software engineers). Hence, the majority of the people worldwide also don't know this.

I want Flathub to be more transparent/clear on this matter!

What are the rules:
- one whitespace between the value and the unit;
- first letter is always in uppercase (`B`, `K`, `M`, `G`, `T`, `P`, `E`, `Z`, `Y`);
- if value is `>= 1024` then:
  - the second letter is `i` (lowercase; from "binary");
  - the third letter is `B` (uppercase; from "byte").

See https://en.wikipedia.org/wiki/Binary_prefix

Andrew15-5:
I tidied up the code (from the first commit). Things to note:
- added explicit whitespace (`" "`) in string concatenation
- added explicit const for `"i"` for use in string concatenation
- added alias for 1024 (because it's used several times)
- if only 1 value is converted to string, then used more readable `.toString()` in string concatenation
- the function runs in O(1) using log with base 1024 to get the power, but not more than `UNIT_CHARS.length - 1` (because YiB is the last unit which is `2^10^8`)
- I think I followed all the styling conventions from this project (I also used `prettier`)

Closes #2139.

### DIFF
--- a/frontend/src/size.ts
+++ b/frontend/src/size.ts
@@ -1,9 +1,31 @@
+function roundNumber(number: number, digits: number) {
+  const multiple = Math.pow(10, digits)
+  return Math.floor(number * multiple + 0.50000000000001) / multiple
+}
+
 export function calculateHumanReadableSize(size: number) {
-  if (size > 1_000_000_000) {
-    return (size / 1_000_000_000).toFixed(2) + " GB"
-  } else if (size > 1_000_000) {
-    return Math.round(size / 1_000_000) + " MB"
-  } else {
-    return Math.round(size / 1_000) + " KB"
+  // (Maximum) number of digits to round to after floating point.
+  const ROUND_DIGITS = 2
+
+  // First letters of multiples of 2, more precisely of 2^10.
+  const UNIT_LETTERS = ["B", "K", "M", "G", "T", "P", "E", "Z", "Y"]
+
+  const i = "i" // From "binary".
+  const POW10 = 1024 // 2 to the power of 10.
+
+  if (size < POW10) {
+    // Explicit whitespace for good measures.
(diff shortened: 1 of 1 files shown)


razzeee:
Thanks for looking into this

Andrew15-5:
@razzeee, could you please tell me when the main branch is released into production? Is there a constant interval or is it done on-demand?

razzeee:
It should start to show up as soon as the builds are done, but it sometimes takes some more time to get cleared out of the different caches.

So wait for https://github.com/flathub/website/actions and then some more :)

PR #2142, opened 2023-10-24
Remove vending useAsync calls and remove it all together

razzeee:
@kymckay Can you test this, I've done this mostly blind now, due to not having a vending setup here.

I think we'll also need to talk about `vending.ts`, as it seems it had some logic e.g. `DEFAULT_SETUP`, that should be part of the backend and not in the client.


frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenCancelButton.tsx
```diff
@@ -20,43 +20,41 @@ const TokenCancelButton: FunctionComponent<Props> = ({
...
+    },
+    onSuccess: (data) => {
+      if (data.data[0].status === "cancelled") {
+        setState("cancelled")
+        toast.success(t("token-cancelled"))
+      }
+    },
+    onError: (error) => {
+      toast.error(t(error as string))
+    },
```
kymckay, line 41: Will give this PR a test tomorrow morning and just merge if it's good :+1:

General note: I believe `onSuccess` and `onError` are deprecated. Same effects can be achieved inside the query function.

razzeee: Yes, they are deprecated, but I wasn't sure if that applies to mutations too. But it probably does.

razzeee: Also, we might want to address the `DEFAULT_SETUP` problem first.

razzeee: Then again, this isn't used, so creating an issue for that would also be fine in my book.

razzeee:
Working on this, also made it obvious, that the created names for the endpoints are horrible. I can only figure out what is what by knowing if it's post or get and looking at the params, because the names are soooo cryptic and almost have everything in them.

kymckay:
Got distracted, but testing this now

razzeee:
Followup in https://github.com/flathub/website/pull/2202

kymckay:
Annoyingly Stripe have added additional steps to Express account creation which serve as blockers in test mode 
image

Making payments seem to be working though. I'm getting `invalid-appid` errors when making tokens and accessing payment settings - which may be related to the lack of express account (we should be handling that better regardless).

razzeee:
Can you test if you run into these without these changes?

PR #2143, opened 2023-10-25
Migrate quality moderation to use codegen api

razzeee:
I'm not sure how to correctly type the last remaining endpoint, as the GUIDELINES type seems to not be what pydantic wants us to do


razzeee:
Figured it out

PR #2184, opened 2023-11-04
Update to new major nextjs version

razzeee:


razzeee:
@barthalion this seems to need node 18, do you know which one the server is using, as I think the`Dockerfile` here is only used for dev

PR #2202, opened 2023-11-11
Remove vending useAsync calls and remove it all together

razzeee:


backend/app/vending/__init__.py
```diff
@@ -226,7 +226,9 @@ def status(login=Depends(login_state)) -> VendingStatus:
 
 
 @router.post("/status/onboarding", tags=["vending"])
-def start_onboarding(data: VendingOnboardingRequest, login=Depends(login_state)):
+def start_onboarding(
```
kymckay, line 229: GitHub prevents me commenting on unchanged lines, but on L207 of this file, the return type should match what is declared in the function header. It's currently causing issues with vending link retrieval in the frontend (a query condition falsely becomes true and tries to proceed to dashboard link retrieval).

kymckay (changes requested):
Some routes are giving an `invalid-appid` error:
- `/vendingapp/org.flatpak.qtdemo/setup`
- `/vendingapp/org.flatpak.qtdemo/tokens`

backend/app/vending/__init__.py
kymckay: I fixed in testing locally by just returning a default `VendingStatus` with `status="none"` or similar and all `False` values.

Was unable to push to your fork branch.

frontend/src/components/user/VendingLink.tsx
```diff
@@ -60,13 +67,24 @@ const VendingLink: FunctionComponent = () => {
   }
 
   if (statusQuery.isError || dashboardQuery.isError) {
-    return <p className="m-0">{t(statusQuery.error || dashboardQuery.error)}</p>
+    return (
+      <p className="m-0">
+        {t((statusQuery.error as string) || (dashboardQuery.error as string))}
```
kymckay, line 72:
```suggestion
        {t((statusQuery.error || dashboardQuery.error) as string)}
```

nit

frontend/src/components/user/VendingLink.tsx
```diff
@@ -23,32 +18,44 @@ const VendingLink: FunctionComponent = () => {
...
+    queryFn: () =>
+      vendingApi.getDashboardLinkVendingStatusDashboardlinkGet({
+        withCredentials: true,
+      }),
 
-  // Otherwise onboarding required still
-  // Can't pre-fetch link, it will create an account without user action
-  const startOnboard = useCallback(async () => {
-    if (onboarding) return
+    enabled: statusQuery.isSuccess && statusQuery.data.data.details_submitted,
```
kymckay, line 37: This is the condition which is causing errors in the frontend (becomes true when it shouldn't), which is interesting because it doesn't happen on `main` even though the logic looks identical apart from the extra `.data` attribute in the chain.

kymckay (commented):
Further common on my initial review around the `invalid-appid`: the errors don't occur on `main` and the settings pages behave as expected

backend/app/vending/__init__.py
razzeee: I really don't like that status is just a string :/

razzeee: But changed it

frontend/src/components/user/VendingLink.tsx
razzeee: Well I guess that was caused by the 201 and us then trying to access `details_submitted`?

frontend/src/components/application/AppVendingControls/SetupControls.tsx
```diff
@@ -37,11 +37,10 @@ const SetupControls: FunctionComponent<Props> = ({ app, vendingConfig }) => {
...
 
   // Need existing app vending configuration to initialise controls
-  const {
-    status,
-    value: vendingSetup,
-    error,
-  } = useAsync(useCallback(() => getAppVendingSetup(app.id), [app.id]))
+  const vendingSetup = useQuery({
+    queryKey: ["appVendingSetup", app.id],
+    queryFn: () => vendingApi.getAppVendingSetupVendingappAppIdSetupGet(app.id),
```
kymckay, line 42: Credentials here is the last thing needed and then this LGTM

kymckay (approved):
Approved to unblock merge, one last change needed

frontend/src/components/application/AppVendingControls/SetupControls.tsx
razzeee: Should be fixed now

PR #2227, opened 2023-11-20
Store data in database and allow for different configs

razzeee:
I'm sorry - this is a bit mad, but it seems to work


backend/app/stats.py
```diff
@@ -254,6 +254,13 @@ def update():
             [i[2] for i in dict.values()]
         )
 
+    [
+        models.Apps.set_downloads(
```
razzeee, line 258: Change this to a normal loop

backend/app/models.py
```diff
@@ -1345,13 +1363,50 @@ class ModerationRequest(Base):
     comment = mapped_column(String)
 
 
+class GuidelineCategory(Base):
+    """A category of quality guidelines for an app"""
+
+    __tablename__ = "guideline_category"
```
razzeee, line 1369: This is kind of against the naming of the other tables, but the previous naming scheme makes everything way too hard to read.

razzeee:
@barthalion this should be good to go now. There might be a timeframe between applying the migration and the `/update` running, where some moderator specific pages will not work.

Would be nice to have a backup of `qualitymoderation` before this, just in case. We also might want to watch if the migration runs without issues, as our local env have less historical data, which might cause problems.

PR #2266, opened 2023-12-04
Generate badge on the fly

razzeee:


razzeee:
This now generates a new badge page (badges-new) which isn't linked. But should be safe to merge and then figure out how we need to handle the api endpoint on the nextjs side of things.

frontend/pages/api/badge.tsx
```diff
@@ -0,0 +1,183 @@
...
+) {
+  ;(async () => {
+    const wasmPath = path.join(
+      process.cwd(),
+      "node_modules/@resvg/resvg-wasm/index_bg.wasm",
+    )
+    const wasmBuffer = await fs.readFile(wasmPath)
+    const wasmArrayBuffer = new Uint8Array(wasmBuffer).buffer
+    try {
+      await resvg.initWasm(wasmArrayBuffer)
```
razzeee, line 48: This part is complaining, that it get's called multiple times

PR #2279, opened 2023-12-09
Fix minor error in German translation

bk138:


razzeee:
Should be fixed in the next round of translations :+1:

PR #2281, opened 2023-12-10
Show moderation pending to app authors

razzeee:
FYI @barthalion


razzeee:
This change should be fine, as we already hold stuff for moderation and the email points at the web interface for it. Which some app authors try to find and don't see.

PR #2301, opened 2023-12-13
Add domain name tests and allow subdomains

razzeee:
We might not want this and there could be a security argument made for not enabling it.

Reported via https://discourse.flathub.org/t/verification-for-apps-hosted-on-framagit/5683

### DIFF
--- a/backend/app/verification.py
+++ b/backend/app/verification.py
@@ -134,6 +134,11 @@ def _get_domain_name(app_id: str) -> str | None:
         projectname = _demangle_name(projectname)
         # https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/
         return f"{projectname}.{domain}.io".lower()
+    elif len(app_id.split(".")) == 4:
+        [tld, domain, subdomain] = app_id.split(".")[0:3]
+        subdomain = _demangle_name(subdomain)
+        domain = _demangle_name(domain)
+        return f"{subdomain}.{domain}.{tld}".lower()
     else:
         [tld, domain] = app_id.split(".")[0:2]
         domain = _demangle_name(domain)
--- a/backend/tests/main.py
+++ b/backend/tests/main.py
@@ -465,6 +465,9 @@ def test_verification_domain_names():
         _get_domain_name("net.sourceforge._0example.App") == "0example.sourceforge.io"
     )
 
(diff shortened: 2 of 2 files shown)


razzeee:
I guess stuff like `example.co.uk` would also cause problems and that combined with subdomains might be even worse

nandedamana (commented):
Wouldn't be able to handle `subdomain.example.co.in`

razzeee:
> EDIT: I didn't notice @razzeee's comment; it's the same concern.
> 
> Wouldn't be able to handle `subdomain.example.co.in`

That's in part, why this hasn't been merged. There is the posability, that allowing subdomains will cause this to be attackable.

backend/app/verification.py
```diff
@@ -134,6 +134,11 @@ def _get_domain_name(app_id: str) -> str | None:
         projectname = _demangle_name(projectname)
         # https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/
         return f"{projectname}.{domain}.io".lower()
+    elif len(app_id.split(".")) == 4:
+        [tld, domain, subdomain] = app_id.split(".")[0:3]
+        subdomain = _demangle_name(subdomain)
+        domain = _demangle_name(domain)
+        return f"{subdomain}.{domain}.{tld}".lower()
```
hfiguiere, line 141: the split should be done in reverse. In flathub we have strict policy of the last component being the app name, the rest being a FQDN in reverse.

Many countries have sub domain in the tld (`.jp`, `.uk`, `.in`, etc.) that are the top level for registration leading to at least 3 components for a domain name.

I don't see why you have to split at 3.

razzeee: I don't remember the details, but I think my hope was, to separate the subdomain issue from the tld issue and thus maybe make it easier to get a first change in.

hfiguiere: this https://github.com/flathub-infra/website/pull/3265 looks like the better approach.

razzeee: This is only the "better" version cause it has tests

hfiguiere: test on a broken feature is still a broken feature.

razzeee: not sure that's news

PR #2360, opened 2024-01-01
Allow framagit projects to be verified

danialbehzadi:
This will fix [this bug](https://discourse.flathub.org/t/5683)


razzeee:
No tests?

PR #2362, opened 2024-01-02
Remove project groups

barthalion:


frontend/.env.development
```diff
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_API_BASE_URI=http://localhost:8000
-NEXT_PUBLIC_SITE_BASE_URI=http://localhost:3000
-NEXT_PUBLIC_MATOMO_WEBSITE_ID=38
+NEXT_PUBLIC_API_BASE_URI=https://flathub.org/api/v2
```
razzeee, line 1: This is probably an oversight

barthalion: Yeah, had this in my local checkout. I will remove that.

PR #2442, opened 2024-01-17
moderation: Create GitHub rejection issue for unverified apps

barthalion:


backend/app/emails.py
```diff
@@ -46,6 +46,7 @@ class EmailInfo(BaseModel):
     template_data: dict[str, Any]
     # Only works when app_id is set and email is not user specific
     inform_moderators: bool = False
+    inform_only_moderators: bool = False
```
razzeee, line 49: Do we really want the four cases this generates in combination with `inform_moderators`

I would think we only want three?
- Users
- Moderators
- All

would  `inform_moderators = false` and `inform_only_moderators = true` even make sense?

backend/app/moderation.py
```diff
@@ -459,17 +500,28 @@ def submit_review(
...
+    verification_status = get_verification_status(request.appid)
+
+    if request.is_approved:
+        category = EmailCategory.MODERATION_APPROVED
+        subject = f"Change in build #{request.build_id} approved"
+    else:
+        category = EmailCategory.MODERATION_REJECTED
+        subject = f"Change in build #{request.build_id} rejected"
+
+    if not verification_status.verified:
```
razzeee, line 513: I'm not sure why this is the deciding factor here, as an app can still have developers with emails assigned, even if it's unverified.

backend/app/emails.py
barthalion: We should probably rework this to always send specific types of messages to moderators directly in the e-mail function but I will be worrying about this later

backend/app/moderation.py
barthalion: Marking as resolved as per what I sent on Matrix:

>people who submitted an app using github are more likely to never log in to the website
>so we don't have their e-mails, and they won't know a build was rejected
>while for direct upload, you have to log in to verify the ID before uploading anything

PR #2459, opened 2024-01-20
Implement archival and enrollment to direct uploads

barthalion:


backend/app/moderation.py
```diff
@@ -522,7 +520,7 @@ def submit_review(
         category = EmailCategory.MODERATION_REJECTED
         subject = f"Change in build #{request.build_id} rejected"
 
-        if not verification_status.verified:
+        if models.DirectUploadApp.by_app_id(sqldb, request.appid) is not None:
```
razzeee, line 523: Check for an empty list https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Result.all

barthalion: Yeah, that's some brain fart from programming in the early morning. Fixed

backend/app/logins.py
```diff
@@ -924,6 +926,7 @@ def get_userinfo(login: LoginStatusDep) -> UserInfo:
     invited_flatpaks = [
         app.app_id
         for _invite, app in models.DirectUploadAppInvite.by_developer(db, user)
+        if not app.archived
```
razzeee, line 929: I think we should just default to filtering this on the model. This is also used for the `dev_flatpaks` listing for e.g. and like this it will still show archived repos

backend/app/verification.py
```diff
@@ -252,6 +268,17 @@ class VerificationStatus(BaseModel):
     detail: str | None = None
 
 
+class ArchiveRequest(BaseModel):
+    endoflife: str | None = None
```
razzeee, line 272: I would like this as `endoflife_message`, as I always get confused

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+    login=Depends(logged_in),
+    app_id: str = Path(
+        min_length=6,
+        max_length=255,
+        pattern=r"^[A-Za-z_][\w\-\.]+$",
+        examples=["org.gnome.Glade"],
+    ),
+):
+    if not config.settings.flat_manager_api:
+        return
```
razzeee, line 948: Wondering if this should fail via a raise instead

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+
+    direct_upload_app = models.DirectUploadApp.by_app_id(sqldb, app_id)
+    if direct_upload_app is not None:
+        if direct_upload_app.archived:
+            return
+
+        direct_upload_app.archived = True
+        sqldb.session.commit()
+
+    if not direct_upload_app:
```
razzeee, line 958: Kind of confused, why this is here and not the same as 951 - but it's late :)

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+        gh_repo_changed = _archive_github_repo(app_id)
+        if not gh_repo_changed:
+            return
+
+    if app_id not in login.user.dev_flatpaks(sqldb):
```
razzeee, line 963: This is waaaaaay too late, needs to be the first thing to check

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+            headers={"Authorization": flat_manager_jwt},
+            json={"token_ids": [_jti(token)]},
+        )
+
+        if not response.ok:
+            raise HTTPException(status_code=500)
+        token.revoked = True
+        sqldb.session.commit()
+
+    worker.republish_app.send(app_id, request.endoflife, request.endoflife_rebase)
```
razzeee, line 984: This technically allows to end of life your app and rebase it onto somebody elses app. Do we want to allow that?

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+    )
+    for token in upload_tokens:
+        response = requests.post(
+            config.settings.flat_manager_api + "/api/v1/tokens/revoke",
+            headers={"Authorization": flat_manager_jwt},
+            json={"token_ids": [_jti(token)]},
+        )
+
+        if not response.ok:
+            raise HTTPException(status_code=500)
```
razzeee, line 980: A bit worried about this and other return paths, as you potentially exit early and can't do this again - as you've already commited database things above and that will block you from rerunning.

Or one token failing to be revoked would lead to no republish

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
         worker.republish_app.send(app_id)
 
 
+@router.post("/{app_id}/enroll", status_code=204, tags=["verification"])
+def enroll(
+    login=Depends(logged_in),
```
razzeee, line 917: This seems basic, shouldn't we check if the user actually is the dev of this app?

You can use `Depends(app_author_only)` for that btw

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
         worker.republish_app.send(app_id)
 
 
+@router.post("/{app_id}/enroll", status_code=204, tags=["verification"])
```
razzeee, line 915: I'm struggling with the name here - I guess this is the endpoint for switching from a github repo to a self managed one?

backend/app/verification.py
```diff
@@ -884,6 +912,78 @@ def unverify(
...
+def enroll(
+    login=Depends(logged_in),
+    app_id: str = Path(
+        min_length=6,
+        max_length=255,
+        pattern=r"^[A-Za-z_][\w\-\.]+$",
+        examples=["org.gnome.Glade"],
+    ),
+):
+    is_direct_upload = models.DirectUploadApp.by_app_id(sqldb, app_id) is not None
```
razzeee, line 925: I'm failing a bit to figuring this out.

As far as I know, a direct upload app can't be created with the same `app_id` as an already existing app. So this and having a github repo to archive is impossible?

backend/app/verification.py
```diff
@@ -851,7 +878,8 @@ def confirm_website_verification(
 
         sqldb.session.commit()
 
-        worker.republish_app.send(app_id)
+        if not new_app:
+            worker.republish_app.send(app_id)
```
razzeee, line 882: I guess `verify_by_login_provider` will need this too

backend/app/verification.py
razzeee: Also not sure, why this can be `None` might be a nice thing to force having this.

backend/app/verification.py
barthalion: We already allow that by using flathub.json.

backend/app/verification.py
barthalion: Yeah, I think I was writing it too early in the morning.

backend/app/verification.py
barthalion: I think I need to push this to the worker, and we should also wrap all external requests to retry on failure. Will reshuffle the code but will address this in another PR.

backend/app/verification.py
barthalion: But endpoints are public, ale people usually do dumb things when they can, so it's better safe than sorry here.

backend/app/verification.py
barthalion: Done.

backend/app/logins.py
barthalion: Not super sure if it's not going to break something else, but sure, we can fix it later

backend/app/verification.py
barthalion: Direct upload vs GitHub app, but I guess I expressed conditions in a confusing way. Changed it

backend/app/verification.py
razzeee: I guess what I meant was an api check - and `_create_direct_upload_app` would not fail due to an already existing github app, as we don't enforce that in the database - still slightly worried about a directupload with the same appid as a githubrepo, but I guess that's where the archived flag comes into play.

PR #2461, opened 2024-01-20
Allow transliterate names for non-Latin scripts.

danialbehzadi:
In locales with non-Latin script, the names should be transliterated to be readable.

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -74,7 +74,7 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
       <ThemeProvider attribute="class">
         <DefaultSeo
           dangerouslySetAllPagesToNoIndex={!IS_PRODUCTION}
-          titleTemplate="%s | Flathub"
+          titleTemplate="%s | {t('Flathub')}"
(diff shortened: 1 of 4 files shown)


razzeee:
That needs way more changes, especially for the flathub variable. At least each key needs to be created in the english translation file, but `flathub` shows up in many places more.

I kept it a static string like this for a reason, still not exactly against you changing it, but it will be more work and likely even more to maintain.

razzeee:
There is https://github.com/flathub-infra/website/pull/3335 for the mastodon part of this, along with upstream discussion, so that part should be removed from here

frontend/pages/_app.tsx
```diff
@@ -74,7 +74,7 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
       <ThemeProvider attribute="class">
         <DefaultSeo
           dangerouslySetAllPagesToNoIndex={!IS_PRODUCTION}
-          titleTemplate="%s | Flathub"
+          titleTemplate="%s | {t('Flathub')}"
```
razzeee, line 77:
```suggestion
          titleTemplate={`%s | ${t("Flathub")}`}
```

The old code doesn't work, it should be fine like this

PR #2468, opened 2024-01-21
Show github link after creation and use modal component

razzeee:


razzeee:
@barthalion the link part needs testing and will only show right after something was rejected

PR #2496, opened 2024-01-24
Improve typing and queryClient state

razzeee:


razzeee:
Code creation doesn't seem to be able to create these types atm

PR #2509, opened 2024-01-26
Use email message class

razzeee:


backend/app/invites.py
```diff
@@ -141,6 +141,7 @@ def invite_developer(
 
     worker.send_email.send(
         EmailInfo(
+            message_id=f"{app.app_id}/{invited_user.id}/invited",
```
hfiguiere, line 144: these message id don't look very unique. Uniqueness is paramount here.

razzeee: the invite cases should now be better

backend/app/moderation.py
```diff
@@ -441,6 +441,7 @@ def submit_review_request(
                 requests = list(requests)
                 worker.send_email.send(
                     EmailInfo(
+                        message_id=f"{app_id}/{review_request.build_id}/held",
```
razzeee, line 444: Worried about `build_id` as that might not be very unique, especially when we get external builds. They are also used in the subject, so that's a problem.

Issue #50, opened 2021-10-03
Implement a review backend


razzeee:

razzeee:
Should wait for https://gitlab.gnome.org/Infrastructure/odrs-web/-/merge_requests/17 so that we can filter for actual reviews of the flatpak versions

sonnyp:
odrs lack features that are important for publishers, such as replying to reviews publicly or privately contacting a review author to ask for more details.

Would these be doable wth odrs?

razzeee:
I'm wondering if just starting with stars would be any improvement.

mijorus:
Hi @razzeee, I would like to contribute to this feature.
  
Is this https://github.com/flathub-infra/website/issues/50#issuecomment-1179700245 this a blocker for you?
I know that mixing reviews from multiple sources is not ideal, but as a user and an app developer it would be great to see all my reviews on a single place.

I basically had to make my [own implementation](https://flatstat.mijorus.it/app/it.mijorus.smile) of that, just to see the reviews of my app

razzeee:
hey,

> Is this https://github.com/flathub-infra/website/issues/50#issuecomment-1179700245 this a blocker for you?

I at least thought so, as I suspect reviews heavily diff between the way something is packaged. 

> I know that mixing reviews from multiple sources is not ideal, but as a user and an app developer it would be great to see all my reviews on a single place.

Not sure why as a `user` - I do get that as a `developer`, but thought you could also just login to odrs for that (I've only patched that thing a few times, never really used it)

mijorus:
I think it makes sense to show reviews and star ratings at least for verified apps, as the developer is actively targeting Flathub, so there shouldn't be issues related to sandboxing.


>  but thought you could also just login to odrs for that 

Can I? Not really sure where or how, their website says I can "request an account if you think it is required", without clarifying where to ask for.

mijorus:
I'll send a pr anyway, then you can decide what you think about it

razzeee:
Please wirte an outline of what you are planning to do first

mijorus:
Sure. This sis my proposal

1. We start by implementing the rating system first, comments and reviews later, as we need to decide how to handle multiple languages
2. Only verified apps should have visible reviews: since the developer is officially publishing the Flathub release, I think we can assume that the app should work flawlessly and is being developed with Flatpak's limitations in mind
3. We can create a backend api that caches ratings from ODRS, say once a day, with redis: the API responds with the average rating already computed
[...]

PR #2510, opened 2024-01-26
Show app ratings from GNOME odrs project

mijorus:
This PR I have implemented some logic to automatically refresh app ratings from ODRS's api, cache the result and return it to the frontent.

As discuss here https://github.com/flathub-infra/website/issues/50#issuecomment-1912522858, this implementation adds app ratings (stars) only for verified apps

Screenshot 2024-01-26 alle 22 21 52

### DIFF
--- a/backend/app/app_reviews.py
+++ b/backend/app/app_reviews.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from datetime import timedelta
+from pydantic import BaseModel
 
 import logging
 import requests
@@ -13,7 +14,10 @@ REVIEWS_CACHE_KEY = 'odrs_reviews'
 @dataclass
 class AppReview: 
     average_rating: float
-    
+
+class LoadRatingsQuery(BaseModel):
+  query: list[str]
+
 def _update_reviews():
     try:
         reviews = requests.get(f'{ODRS_BASE_URL}/ratings', timeout=15).json()
@@ -25,13 +29,13 @@ def _update_reviews():
     return reviews
 
 
-def get_reviews(ids: list[str]) -> list[AppReview]:
+def get_reviews(ids: list[str]) -> dict[str, AppReview]:
     reviews = db.get_json_key(REVIEWS_CACHE_KEY)
     if not reviews: 
         reviews = _update_reviews()
 
         if not reviews:
-            return []
+            return {}
 
     keys_weight  = {
         "star0": 0,
@@ -42,13 +46,14 @@ def get_reviews(ids: list[str]) -> list[AppReview]:
         "star5": 5,
     }
     
-    result = []
+    result = {}
+
     ids_left = ids.copy()
-    for r, review in reviews.items():
+    for app_id, review in reviews.items():
         if not ids_left:
             break
         
-        if r not in ids:
+        if app_id not in ids:
             continue
         
         weight = 0
@@ -57,6 +62,7 @@ def get_reviews(ids: list[str]) -> list[AppReview]:
         
         average_rating = round(weight / review['total'], 1)
         app_review = AppReview(average_rating=average_rating)
-        result.append(app_review)
-        
+        result[app_id] = app_review
+        ids_left.remove(app_id)
+
     return result
--- a/backend/app/app_reviews.py
+++ b/backend/app/app_reviews.py
@@ -1,17 +1,38 @@
 from dataclasses import dataclass
+from datetime import timedelta
+
+import logging
 import requests
+import orjson
+
+from . import db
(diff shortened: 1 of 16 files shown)


bertob:
I wouldn't clutter the app cards with this, but I think it'd be worthwhile to add to the app pages.

mijorus:
> I wouldn't clutter the app cards with this, but I think it'd be worthwhile to add to the app pages.

Yeah I also have the feeling that this solution doesn't look ok on the home page, but if you hide the ratings all together you kind of miss the point of it, which is to make comparisons (and also incentivise users to leave those ratings in the first place)

mijorus:
I tried to move the verification badge and the stars in a new line, but those apps felt "unbalanced", I don't know how to explain it better.

backend/app/apps.py
```diff
@@ -92,6 +94,17 @@ def show_in_frontend(app: dict) -> bool:
...
+    reviews = app_reviews.get_reviews(list(apps.keys()))
+    
+    for app_id in reviews:
+        if not app_id in apps:
+            continue
+
+        if not 'metadata' in apps[app_id]:
+            apps[app_id]['metadata'] = {}
+
+        apps[app_id]['metadata']['flathub::rating'] = reviews[app_id].average_rating
```
barthalion, line 107: This should be a separate endpoint. ODRS ratings are not appstream.

backend/app/app_reviews.py
```diff
@@ -0,0 +1,68 @@
+from dataclasses import dataclass
```
barthalion, line 1: I do have some doubts about using ODRS in general, especially given it's not being developed at this point and doesn't provide a way to filter out snap and distribution packages.

mijorus: We already talked about this here https://github.com/flathub-infra/website/issues/50#issuecomment-1912522858

Let me know what you guys want to do in the end..

bertob:
> you kind of miss the point of it, which is to make comparisons (and also incentivise users to leave those ratings in the first place)

I don't think that's the point of ratings, they're mostly useful for ranking search results and as part of the review system.

mijorus:
> > you kind of miss the point of it, which is to make comparisons (and also incentivise users to leave those ratings in the first place)
> 
> I don't think that's the point of ratings, they're mostly useful for ranking search results and as part of the review system.

Well, that is basically doing comparisons isn't it? But on the server

mijorus:
Any update on this one?

bertob:
> Well, that is basically making comparisons isn't it? But on the server

Sure, but that doesn't require showing them on the app cards.

JakobDev:
>only for verified apps

Why?

Another problem: ODRS doesn't differ how a App is packaged. If a User gives a 1 Star rating because the native package of some Distro is broken, it will also affect the rating on Flathub. On the other side Users may give a good rating for a native package while the Flatpak is broken.

razzeee:
> Another problem: ODRS doesn't differ how a App is packaged. If a User gives a 1 Star rating because the native package of some Distro is broken, it will also affect the rating on Flathub. On the other side Users may give a good rating for a native package while the Flatpak is broken.

We talked about this in the previous issue

razzeee:
> I feel this is fine to display in the developer portal, but without full ODRS rework, we should not display it on the website.

I'm wondering, if that's even worth the hassle then? My main motivation to get this on the page, would be to be able to use it as SEO data and maybe get search engines to show stars for apps.

Why would this be useful for app authors themselves?

mijorus:
> > I feel this is fine to display in the developer portal, but without full ODRS rework, we should not display it on the website.
> 
> I'm wondering, if that's even worth the hassle then? My main motivation to get this on the page, would be to be able to use it as SEO data and maybe get search engines to show stars for apps.
> 
> Why would this be useful for app authors themselves?

Personally I find ratings and reviews very useful, it's nice to hear feedbacks from users that don't have GitHub or just don't want to open a bug report, maybe because they have a positive experience.

razzeee:
> > > I feel this is fine to display in the developer portal, but without full ODRS rework, we should not display it on the website.
> > 
> > 
> > I'm wondering, if that's even worth the hassle then? My main motivation to get this on the page, would be to be able to use it as SEO data and maybe get search engines to show stars for apps.
> > Why would this be useful for app authors themselves?
> 
> Personally I find ratings and reviews very useful, it's nice to hear feedbacks from users that don't have GitHub or just don't want to open a bug report, maybe because they have a positive experience.

Right, forgot, that we don't just get the average rating, but also the raw rating data and showing these would for sure be valuable to developers.

PR #2513, opened 2024-01-27
Bump next to 14.1, add Redis-based cache

barthalion:


better-salmon:
Hello @barthalion! I am a maintainer of the [@neshca/cache-handler](https://caching-tools.github.io/next-shared-cache/) library. I noticed that you reverted the commit with the library. Could you please tell me what went wrong? Maybe I can help you.

barthalion:
Hey @better-salmon, no idea how you found us :see_no_evil: We've had some issues with the setup pages like https://flathub.org/setup/Ubuntu returning 404; I only noticed when I purged CDN cache, and as it turns out, we've been serving a stale page from there all along. I suspect some bug in the code unrelated to `@neshca/cache-handler` itself, but need to wrap up some other work before returning to this.

Another issue, although unrelated to the revert, was caching "negative hits" without TTL, I ended up using another example from the documentation at https://github.com/flathub-infra/website/commit/63a84865f9cbb9d76b3c5b9dc9b85ae84006a99f but with expiry hardcoded to 1800.

barthalion:
Yes, that would be super helpful!

PR #2514, opened 2024-01-28
Start an email service

razzeee:


razzeee:
I've started to test react-email in some other context, turns out these don't render correctly in outlook. We probably don't care. And there is word of a new outlook renderer coming somewhen.

PR #2524, opened 2024-01-30
Add Redis cache handler for ISR and SSG pages

better-salmon:
This pull request adds a Redis cache handler for Incremental Static Regeneration (ISR) and Static Site Generation (SSG) pages.

I have some concerns regarding this:

- According to `backend/README.md`, the Redis Server URL is `redis://localhost:6379`, but it may only apply to the backend environment.
- The custom Redis Handler assumes that the Redis Server is already reachable when building a Next.js app. If we want it to connect to the Redis Server only at the start of the Next.js app, we need to add some extra conditions in the `IncrementalCache.onCreation` callback.
- We should consider adding an [Eviction policy](https://redis.io/docs/reference/eviction/) because Redis may run out of memory.
- Is it possible to rebuild and redeploy the frontend while preserving the backend (and Redis Server)? If this is the case, we have to use [prefixes](https://www.oreilly.com/library/view/redis-essentials/9781784392451/ch06s03.html#:~:text=Redis%20does%20not%20support%20namespacing,prefix%20to%20all%20Redis%20keys.) or [databases](https://www.digitalocean.com/community/cheatsheets/how-to-manage-redis-databases-and-keys) and flush them on the frontend deployment.


better-salmon:
@barthalion hey! I've created a PR as I promised. Please pay attention to my concerns in a description.

#2513 for a context.

razzeee:
This might also need the nextjs bump?

frontend/cache-handler.js
```diff
@@ -0,0 +1,34 @@
+const { IncrementalCache } = require("@neshca/cache-handler")
```
razzeee, line 1: If this has to be a js file, can we at least get a `// @ts-check` annotation?

frontend/cache-handler.js
```diff
@@ -0,0 +1,34 @@
...
+const createRedisHandler =
+  require("@neshca/cache-handler/redis-strings").default
+const createLruHandler = require("@neshca/cache-handler/local-lru").default
+const { createClient } = require("redis")
+
+const client = createClient({
+  url: "redis://localhost:6379", // based on backend/README.md
```
razzeee, line 8:
```suggestion
url: process.env.REDIS_URL ?? "redis://localhost:6379",
```

This was used by the previous work, so should work fine

razzeee:
> Is it possible to rebuild and redeploy the frontend while preserving the backend (and Redis Server)? If this is the case, we have to use [prefixes](https://www.oreilly.com/library/view/redis-essentials/9781784392451/ch06s03.html#:~:text=Redis%20does%20not%20support%20namespacing,prefix%20to%20all%20Redis%20keys.) or [databases](https://www.digitalocean.com/community/cheatsheets/how-to-manage-redis-databases-and-keys) and flush them on the frontend deployment.

Yeah, there is other data in redis already, that we don't want to drop

frontend/cache-handler.js
better-salmon: Done

frontend/cache-handler.js
better-salmon: Done

better-salmon:
> Yeah, there is other data in redis already, that we don't want to drop

I have added a `keyPrefix` consisting of the static prefix `nextjs-cache:` and a Next.js build ID. You can use these to flush all cache-related keys or keys related to specific frontend builds.

How to find and clear cache if needed:

```bash
redis-cli --scan --pattern nextjs-cache:* | xargs redis-cli del
```

```bash
redis-cli --scan --pattern nextjs-cache:id:* | xargs redis-cli del
```

barthalion:
I will get back to this around Friday. FWIW, we have a separate Redis instance just for the frontend.

better-salmon:
I'm glad to hear that! Feel free to tag me in issues or PRs if you have any questions or issues related to caching or with the caching library.

razzeee:
hey @better-salmon hope you're doing fine

we still seem to see some weird behavior, maybe you have an idea on why.

Sometimes when navigating a page will return a 404, but when you open it in a new tab, it's fine. Currently that's happening with `Mixxx` on the start page for example, but I've seen it with various apps from time to time.

The prefetch next.js does also seems to fail with a 404 
image

better-salmon:
@razzeee Hi there! I'm doing well, thank you. Could you please share the link to the code for those pages?

better-salmon:
Do you notice this problem occurring when you deploy a new frontend/backend release?

razzeee:
Do you mean if it goes away then? I'm unsure, previously @barthalion could drop the cache folder on the server, but that was always a very bad solution.

better-salmon:
TLDR: Preserving a cache between releases may cause such problems.

Let's forget about CacheHandler with an external store for a moment. Typically, you may see the prefetch 404 errors when you release a new version of your Next.js app, while in the browser, you have the old version running, and you click on any link. The process is the following:

1. The user clicks on the link with the old resource
2. New version of your Next.js server responses with 404
3. The server and the client will trigger a full page reload to update all resources listed in HTML head and body
4. The user sees the new version

When you introduce CacheHandler with an external store, you may run into a situation where the old version of the HTML file is retrieved from the cache. It only happens if you keep a cache between the frontend releases. Currently, there are two possible solutions:

1. Purge cache on frontend release
2. Use build ID for cache key prefixing

The second one is the easiest, but it is space-inefficient.

I'm finishing the work on the 1.0.0 version of the CacheHandler, where TTL will work both for the App and Pages Router. And this option will become dramatically more space efficient. For now, it works only for the App Router. I will release it within a week. But for now, I recommend using a build ID for keys or crafting a webhook to purge the frontend release's cache.

https://caching-tools.github.io/next-shared-cache/configuration/build-id-as-prefix-key

razzeee:
Thank you, we have been living with this for a while, so we're fine to wait

better-salmon:
@razzeee, @barthalion hello! As I promised I've released the new version of CacheHandler and made a PR #2652 to your project.

PR #2563, opened 2024-02-07
Update Mastodon url

sonnyp:
https://floss.social/@flathub

Should also take care of verification thanks to `rel=me`


cassidyjames (approved):
This should do it! Thanks for the PR.

PR #2574, opened 2024-02-10
Rework security labels for devices

sonnyp:
"can-access-hardware-devices" is incorrect for `device=all` since it allows access to emulated/"software" devices as well

The label for `can-access-input-devices` is incorrect. `--device=input` doesn't allow access to keyboard and mices. See https://gitlab.gnome.org/GNOME/epiphany/-/merge_requests/1428#note_1993226
It's mostly useful for gamepads and usb gadgets


frontend/public/locales/en/common.json
```diff
@@ -402,8 +402,8 @@
   "has-network-access": "Has network access",
   "uses-system-services": "Uses system services",
   "uses-session-services": "Uses session services",
-  "can-access-hardware-devices": "Can access hardware devices such as webcams or gaming controllers",
-  "can-access-input-devices": "Can access input devices such as keyboards, mice or gaming controllers",
+  "can-access-all-devices": "Can access all devices such as webcams, microphones or keyboards",
```
bertob, line 405:
```suggestion
  "can-access-all-devices": "Can access all devices such as webcams, microphones, or keyboards",
```

razzeee:
I think this will blow up the weblate repo, I usually try to only edit the main translation `en` - not sure renaming complete keys won't cause a merge conflict in their repo.

PR #2587, opened 2024-02-13
Make titles use consistent case on home page

kra-mo:


razzeee:
Please only change the en.json

I would also prefer to have the categories changed instead of the strings you touched now.

kra-mo:
What do you mean by "the categories"?

kra-mo:
Yeah, would you change the appstream ones to not be title case then?

kra-mo:
That just wouldn't look very good because of the &'s separating what are two identically weighted categories.

image

kra-mo:
Okay, I have reverted en-GB

razzeee:
Thank you.

Personally I would prefer the other way around, but as the main headline is using the same casing, whatever :)

The home page will hopefully see a rework soonish anyway.

PR #2592, opened 2024-02-14
utils: Adapt to libappstream

barthalion:


backend/app/utils.py
```diff
@@ -273,6 +265,11 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
         if app["is_free_license"] == "":
             app["is_free_license"] = False
 
+        # The new appstream spec uses a new <developer> key, so backfill the old
+        # field for backwards compatibility
+        if developer_name := app.get("developer", {}).get("name"):
```
razzeee, line 270: I think `developer` can be a list too? So this might fail in those cases?

backend/tests/results/test_apps.json
```diff
@@ -18,6 +18,6 @@
     "name": "WPS Office",
     "summary": "WPS Office Suite",
     "type": "desktop-application",
-    "icon": "https://www.wps.com/assets/mediahub/WPS-Office-Mobile.png"
+    "icon": null
```
razzeee, line 21: This is due to remote icons being a no go now?

barthalion: Yes, apps with remote icons built with appstream-glib will no longer have this defined. The ones that used libappstream will have this properly mirrored to our infra

backend/app/utils.py
barthalion: Hmm, possibly. Can fix that later.

PR #2605, opened 2024-02-19
Add branding color quality check

razzeee:


backend/alembic/versions/248a867312b4_add_branding_quality_guideline.py
```diff
@@ -0,0 +1,54 @@
...
+            sa.Column("needed_to_pass_since", sa.Date(), nullable=False),
+            sa.Column("read_only", sa.Boolean(), nullable=False),
+            sa.Column("guideline_category_id", sa.String(), nullable=False),
+            sa.Column("show_on_fullscreen_app", sa.Boolean(), nullable=False),
+            sa.Column("order", sa.Integer(), nullable=False),
+        ),
+        [
+            {
+                "id": "general-has-primary-branding",
+                "url": "https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#brand-color",
```
razzeee, line 39: We likely want to redo the docs here and maybe change the headline and thus this link

bertob: "Brand color" sounds fine to me :+1:

razzeee: I think it should be more self explanatory? It feels a bit displaced next to the other ones

razzeee: More like `has primary branding colors` (similar to `At least one screenshots`, that's the vibe I'm after)

razzeee: I think the appstream docs touch on that briefly, but we could too

PR #2631, opened 2024-02-25
utils: developer/name child tag is also a list

bbhtt:
Extract only the source tag and ignore translations

### DIFF
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -229,8 +229,12 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
         if len(developers):
             app["developers"] = []
             for developer in developers:
-                developer_name = developer.find("name")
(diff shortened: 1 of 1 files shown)


bbhtt:
jfyi, developer cannot be a list, only one tag is allowed. I don't know how multiple ids are supposed to be handled but that's the behaviour right now

bbhtt:
Looks like someone already opened an issue about the first one https://github.com/ximion/appstream/issues/606

I'll open one for the second

bbhtt:
@barthalion Any idea why the new field is being backfilled twice? The first one seems a typo ? Missing a `s`

https://github.com/flathub-infra/website/blob/21fcc13dd43df61a156f9af1823948e3317cf087/backend/app/utils.py#L289-L303

PR #2634, opened 2024-02-25
Fall back to generic

razzeee:


bbhtt:
I set it in the parsing code https://github.com/flathub-infra/website/pull/2633 I guess there's no difference between here and there?

razzeee:
Yours is probably better, we should really type this.

PR #2646, opened 2024-02-27
Separate developer portal from my flathub

razzeee:
This separates the developer facing bits into a developer portal and leaves the my flathub page as the page for normal users.

We currently don't have any actions normal users can do, so show a nice empty page.

The developer portal can be found from the footer.


razzeee:
@cassidyjames I would like some feedback on the strings/wording

frontend/public/locales/en/common.json
```diff
@@ -208,7 +209,7 @@
   "account-deletion-token-mismatch": "The account deletion request failed to authenticate.",
   "privacy-policy": "Privacy Policy",
   "privacy-policy-description": "By using this website, you agree to the Privacy Policy.",
-  "terms-and-conditions": "Terms and conditions",
+  "terms-and-conditions": "Terms And Conditions",
```
cassidyjames, line 212: Traditionally minor words/short conjunctions are not capitalized for title case

```suggestion
  "terms-and-conditions": "Terms and Conditions",
```

cassidyjames:
Copy changes look good to me here; I haven't spun this up to test it locally, though.

PR #2652, opened 2024-02-28
Update cache-handler module and configuration

better-salmon:
This pull request updates the cache-handler package and configuration:

- Updated the `@neshca/cache-handler` package to version 1.0.4
- Removed `@neshca/json-replacer-reviver`
- Added the usage of `buildId` as `keyPrefix`
- Added proper connection failure checks

### Things to consider

The new version of the CacheHandler supports automatic key expiration. It means that cache entries will be removed from the Redis after `revalidate` value from `getStaticProps` plus 60 seconds. The entries with `revalidate: false` will be removed after one year plus 60 seconds. These 60 seconds give the Next.js server the ability to serve staled value to a user once and revalidate it in the background. It is an emulation of the stale-while-revalidate strategy.

The only exceptions are the pre-rendered pages without `getStaticPaths` or with `fallback: false` in `getStaticPaths`. Such entries have no auto-expiration by design. They needed to be removed manually to free space.

As a result, automatic key expiration improves space efficiency in the Redis database, and now it is possible to use `keyPrefix` to deal with [collisions when preserving a Redis database between frontend releases](https://github.com/flathub-infra/website/pull/2524#issuecomment-1960318305).

To configure stale-while-revalidate strategy emulation you should use `estimateExpireAge` option.

To make `buildId` deterministic see [`generateBuildId`](https://nextjs.org/docs/app/building-your-application/deploying#build-cache) `next.config.js` option.


razzeee:
I guess we would want

```
module.exports = {
  generateBuildId: async () => {
    // This could be anything, using the latest git hash
    return process.env.GITHUB_SHA
  },
}
```

then - not sure if we need to pass manually to from the pipeline

better-salmon:
I've added `GITHUB_SHA` to workflows and used it in the `next.config.js`

better-salmon:
I'm sorry for that. I oversaw the export statement of `cache-handler.mjs` module. I made a fix #2659.

I have also improved `buildId` generation logic for local and non GitHub environments.

PR #2654, opened 2024-02-29
Add roles and rights tables

razzeee:
@barthalion this is unfinished, but hopefully allows you to judge, if this should work okay


backend/app/emails.py
```diff
@@ -136,11 +136,12 @@ def send_email(info: EmailInfo, db):
                     messages.append(_create_message(user, info, db))
 
         if info.inform_only_moderators or info.inform_moderators:
-            moderators = db.session.query(models.FlathubUser).filter_by(
-                is_moderator=True
-            )
-            for user in moderators:
-                messages.append(_create_message(user, info, db))
+            moderator_role = models.FlathubUser.by_role(db, "moderator")
```
razzeee, line 139: We will have to explicitly give the moderator role (in addition to `admin`) for people to get these mails now.

razzeee:
@barthalion this is good to go now

After merging and deploying, you will need to use your database access to connect users with roles in the `flathubuser_role` table.

So far only `moderator` and `quality-moderator` are used for checks, but it would be neat if you can give yourself and me `admin`.

Permission wise `direct-upload` and `payment` is not used yet, I think that's the next step after this went in.

PR #2659, opened 2024-03-03
Fix export statement and update buildId generation logic

better-salmon:
This pull request fixes an export statement in the cache-handler module and updates the buildId generation logic in the next.config.js file.

### DIFF
--- a/frontend/cache-handler.mjs
+++ b/frontend/cache-handler.mjs
@@ -54,4 +54,4 @@ CacheHandler.onCreation(async ({ buildId }) => {
   }
 })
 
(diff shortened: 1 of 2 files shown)


razzeee:
hey, @better-salmon hope your doing good.

I'm a little bit stuck with a problem, maybe you have an idea.
Some of our internal/protected pages, keep 404ing, as they seem to point to an old build cache? (I'm not 100% sure)

image

image

Seems like the html document doesn't get refreshed and still references the old files?

Has this anything to do with the cache or is that a different layer?

PR #2680, opened 2024-03-08
Use new email service

razzeee:


razzeee:
I've completly tested the following email flows:

- login
- moderation
	- approve with commen
	- approve
	- reject

moderation held is quiet hard to test locally and the other emails are not in use yet as far as I know

razzeee:
This has been landed in pieces in different MRs

Issue #2708, opened 2024-03-13
Add Kabyle language


BoFFire:
Hello,

I added Kabyle language via Weblate but the strings appear to be 0. Can you help to create `kab` with empty template and sync that with Weblate to be able to translate?

Screenshot here : https://mstdn.fr/@ButterflyOfFire/112090025162575445

Check this PR where Kabyle is empty : https://github.com/flathub-infra/website/pull/2707/files

Thanks

gunchleoc:
What's your fallback locale for datetime? `ar-DZ`? Full list is here: https://github.com/date-fns/date-fns/tree/main/src/locale

BoFFire:
Hello @gunchleoc 

> What's your fallback locale for datetime? `ar-DZ`? Full list is here: https://github.com/date-fns/date-fns/tree/main/src/locale

Not `ar-DZ` but English `en` or `en-US`should be okay.

gunchleoc:
@BoFFire PR is done, please double-check: https://github.com/flathub-infra/website/pull/2709

I found some more dependencies, in case you want to translate those too, here's the list:

- https://github.com/michaelwittig/node-i18n-iso-countries/tree/master/langs
- https://github.com/i18next/next-i18next (doesn't seem to need translations
- https://github.com/date-fns/date-fns/tree/main/src/locale

BoFFire:
Hi @gunchleoc 
Thank you ! I will translate the other linked projects. I'm adding them to my local translation queue.

razzeee:
You should be able to translate it now, seems to show up fine on weblate

BoFFire:
Thank you @razzeee and @gunchleoc I started with 12% translation.

PR #2709, opened 2024-03-13
Add kabyle locale

gunchleoc:
Closes https://github.com/flathub-infra/website/issues/2708

Note that I don't know how to test this, I just grepped for `zh` and did my best. I can't learn a new codebase right now.

Kabyle language name translation taken from https://www.unicode.org/cldr/charts/44/summary/kab.html#787d4e0576d7869a

The flag is Algeria https://en.wikipedia.org/wiki/Kabyle_language

### DIFF
--- a/frontend/next-i18next.config.js
+++ b/frontend/next-i18next.config.js
@@ -52,6 +52,7 @@ module.exports = {
       "pa",
       "sq",
       "ia",
+      "kab",
     ],
   },
   // debug: true,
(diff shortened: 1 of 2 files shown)


razzeee:
These changes need to be done in addition, after translations become available

BoFFire:
@razzeee Hi,

The problem is that I cannot translate in Kabyle : the file is empty. In Weblate it says 0 strings.

razzeee:
Yeah, usually it sorts itself out somewhen

BoFFire:
Wait and see

PR #2736, opened 2024-03-19
Setup sentry for hono

razzeee:


backend-node/src/index.ts
```diff
@@ -189,6 +190,14 @@ const route = createRoute({
 })
 const app = new OpenAPIHono()
 
+app.use(
+  "*",
+  sentry({
+    dsn: process.env.SENTRY_DSN,
```
razzeee, line 196: We should probably use a different variable name as it should also be a different DSN for the backend

Issue #2525, opened 2024-01-30
The og:image should be larger than 128x128


soyersoyer:
Now the og:image is 128x128:
```
<meta property="og:image" content="https://dl.flathub.org/repo/appstream/x86_64/icons/128x128/io.github.tsmetana.mpk3-settings.png">
```

And this makes a link in reddit ugly:

![Screenshot from 2024-01-30 17-41-16](https://github.com/flathub-infra/website/assets/5169997/823bc5ff-32e3-4fdb-aac6-dd1e1a6c390b)

According to the [best practices](https://developers.facebook.com/docs/sharing/best-practices):

> Images
> 
> Use images that are at least 1080 pixels in width for best display on high resolution devices. At the minimum, you should use images that are 600 pixels in width to display image link ads. We recommend using 1:1 images in your ad creatives for better performance with image link ads.
> 
> Pre-cache your images by running the URL through the [URL Sharing Debugger tool](https://developers.facebook.com/tools/debug) to pre-fetch metadata for the website. You should also do this if you update the image for a piece of content.
> 
> Use og:image:width and og:image:height Open Graph tags to specify the image dimensions to the crawler so that it can render the image immediately without having to asynchronously download and process it.
> 

Probably better to use the first screenshot or a larger icon.

cassidyjames:
We can't guarantee a larger icon exists, and just upscaling a 128px icon would be a no-go. What could be interesting (but would require a bit of setup) is generating an OG image from the name, icon, brand color, and screenshot; basically, save the banner as an image for social media sharing. That would enable really pretty and dynamic images.

cassidyjames:
@razzeee brand color as the background would be nice and make each a bit more distinct; I'd be happy to look into it if you can share how/where you were planning to implement that?

razzeee:
Look at how we generate the badges on the frontend side. Using "satori".

PR #2771, opened 2024-03-23
Generate og images

razzeee:
Generate nicer og images on the fly - this only adds the generation, they are not used yet

image

This defaults to svg for now, as shadows seem broken with png - so we will need to revisit this.

https://github.com/flathub-infra/website/issues/2525


razzeee:
Will wait for barth to return, as it's unclear how much performance this needs from the server and it would be nice to monitor it

bertob:
A few notes based on the screenshot above:

- The logo in the bottom right needs more padding from the edges
- I'd make the title and subtitle a bit larger and reduce the spacing between them
- Ideally I'd make the icon larger as well, but IIRC we discussed this already and the problem was that we can't get a larger asset?

razzeee:
Latest looks like this

image

razzeee:
image

razzeee:
Adjusted logo padding to the right

image

razzeee:
Case without brand color and mobile screenshot

image

bertob:
Vertical spacing looks good now, but I'd bump horizontal spacing a bit to match vertical. Something like 2x the current value should work.

PR #2816, opened 2024-03-30
Use orval instead of openapi-generator-cli to generate api client

razzeee:


razzeee:
This should be good to go, but we might want to have sentry runnig again when we merge this, to be able to see if it breaks anything.

PR #2837, opened 2024-04-02
Only show verified for apps

### LINKED ISSUE #2836: Remove verified/unverified from extensions
We currently have no way to verify extensions. So the badge should be removed. It already upsets people when they feel forced to do it, or when third-party come and file issues about it, so if they can't even address it, it's even worse.

razzeee:
Closes #2836


bbhtt:
The check should probably be inverted here because console-applications get their own page too. You should ignore generic, addon and runtime.

razzeee:
> The check should probably be inverted here/also include console-applications because console-applications get their own page too if they supply the data.

Thank you, we really need to teach the types to the python backend and inherit them into the frontend

PR #2896, opened 2024-04-09
Non-verified applications: Use Unicode warning sign instead of the on…

Alexander-Wilms:
…ne from react-icons

The original one was hard to recognize at this small size

Before:
image

After:
image


razzeee:
Very used to use heroicons directly - this should have used the micro icons, but react-icon doesn't have those. I've created https://github.com/react-icons/react-icons/issues/906 some time ago.

Unfortunatly we actually need react-icons for some icons that we couldn't find something good for in heroicons.

razzeee:
https://github.com/react-icons/react-icons/pull/928

PR #2916, opened 2024-04-11
Add `sudo` before `apt install`

xplosionmind:
Make Flatpak installation instructions straightforward and ready to be copy-pasted: add `sudo` where it is needed.

(`sudo` was already present for Ubuntu instructions, but not in some other cases.)


razzeee:
Hey, we only mirror these from https://github.com/flatpak/flatpak.github.io so you would need to change it there first

xplosionmind:
[Done](https://github.com/flatpak/flatpak.github.io/pull/671). Thanks!

Issue #3156, opened 2024-05-07
Carousel goes too fast when you click next


JamesOsborn-SE:
This does not happen on swipe.

To reproduce:
On the landing page, wait for the carousel to advance, then click next after 4 seconds. After one more second, the carousel will advance again.

Expected behavior:
Autoplay Stops on mouse interaction

PR #3157, opened 2024-05-07
stop autoplay on mouse enter

JamesOsborn-SE:
Addresses issue #3156


razzeee:
Thank you, but I've gone with a slightly different approach, as this approach seemed to have more problems.

PR #3225, opened 2024-05-20
Remember home page section state

### LINKED ISSUE #3200: Home page trending section: Remember state when going back
It's a bit annoying when you want to check out multiple apps in e.g. the "New" section that this part of the page resets when you go back. This happens on both desktop with the tabs, and mobile with the dropdown.

- Open flathub.org
- Click one of the non-default tabs in the new trending/popular/new/updated section, e.g. "New"
- Open one of the apps in that section
- Go back in the browser
- You're back on "Trending"

YonatanPerkel:
Closes #3200

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -124,7 +124,7 @@ const TopSection = ({
       appSelection={
         <>
           <MultiToggle
-            items={topApps.map((x, index) => ({
+            items={topApps.map((x) => ({
               id: x.name,
               content: (
                 <div className="font-semibold truncate">{t(x.name)}</div>
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -137,9 +137,7 @@ const TopSection = ({
                 <div className="font-semibold truncate">{t(x.name)}</div>
               ),
               selected: x.name === selectedName,
-              onClick: () => {
-                setSelectedName(x.name)
-              },
+              onClick: () => setSelectedName(x.name),
             }))}
             size={"lg"}
             variant="secondary"
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -25,6 +25,7 @@ import { AppOfTheDay } from "src/components/application/AppOfTheDay"
 import { formatISO } from "date-fns"
 import { useEffect, useState } from "react"
 import MultiToggle from "src/components/MultiToggle"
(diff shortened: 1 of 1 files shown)


frontend/pages/index.tsx
```diff
@@ -93,12 +94,27 @@ const TopSection = ({
 }) => {
   const { t } = useTranslation()
 
-  const [selectedName, setSelectedName] = useState(topApps[0].name)
+  const [selectedName, setSelectedName] = useState("")
+  const [selectedIndex, setSelectedIndex] = useLocalStorage(
+    "selected-category",
```
razzeee, line 99: Should probably mention, that it only affects the landing page

frontend/pages/index.tsx
```diff
@@ -93,12 +94,27 @@ const TopSection = ({
...
+    0,
```
razzeee, line 100: I would prefer to save the name not the index

YonatanPerkel: Changed local storage to store name instead of index.

frontend/pages/index.tsx
```diff
@@ -109,13 +131,15 @@ const TopSection = ({
       appSelection={
         <>
           <MultiToggle
-            items={topApps.map((x) => ({
+            items={topApps.map((x, index) => ({
```
razzeee, line 134: index can be removed

razzeee:
Thank you, I might need a few days, before I can get back to this and give it it's final testing

razzeee:
This seems to work great.

But I'm unsure, if we really want it like this, as this keeps the state, even if I leave the page and come back days later.

I would much more prefer to always default to trending, if somebody goes directly to flathub.org, but only remember these, when navigating back. I think the "correct" approach for this, would be to place the currently selected item into the url. Similar to how we do it here https://github.com/flathub-infra/website/blob/main/frontend/pages/admin/quality-moderation/index.tsx#L246

YonatanPerkel:
Instead of using url query params, I tried doing it with location hash and I think it's more appropriate for this use case.

razzeee:
> Instead of using url query params, I tried doing it with location hash and I think it's more appropriate for this use case.

I don't feel that way, as it locks us in. There can only be one hash per page - so if we want to have (and remember) a second tab selection on this page, we need to change api and would kill all links, that previously worked.

razzeee:
@YonatanPerkel thank you for looking into this

Issue #3259, opened 2024-05-24
App verification step requires text file under incorrect URL for second-level domains


nandedamana:
I was in the process of verifying my app and the page `https://flathub.org/apps/manage/in.co.nandakumar.vara` showed the incorrect website URL `co.in` instead of `nandakumar.co.in` in these messages:

> Verify your right to use the app ID in.co.nandakumar.vara by placing a token at a specific page on co.in.

> Create a page at https://co.in/.well-known/org.flathub.VerifiedApps.txt containing the following token: 

It's not just the messages; actual verification fails.

nandedamana:
The problem must be here:
https://github.com/flathub-infra/website/blob/e4456e798731a8516881b1cedddb93ff58926d32/backend/app/verification.py#L120

```
...
    else:
        [tld, domain] = app_id.split(".")[0:2]
        domain = _demangle_name(domain)
        return f"{domain}.{tld}".lower()
```

razzeee:
Duplicate of https://github.com/flathub/flathub/issues/4943

hfiguiere:
The rule should be simple. Remove the last bit of the app-id. Then reverse the components.

I don't see why it's hard, this has always been the way. And that how we do during reviews.

We have no way to "manual verify".

nandedamana:
> The rule should be simple. Remove the last bit of the app-id. Then reverse the components.
@hfiguiere just for clarification, you were referring to the process inside the flathub backend, not what I should be doing for fixing the particular package I mentioned, right?

razzeee:
Yes, not on your end.

BTW, there are manual validations, but there is only one person that can do that.

nandedamana:
@razzeee Could you please elaborate on the security concern that you were mentioning in the other threads? Does https://github.com/flathub-infra/website/pull/3265 have the same issue (whatever it is)? I think what I did there was exactly what @hfiguiere mentioned here later.

razzeee:
I think I also played around with python libraries, that support validating by publicsuffix, but I have no idea, if I just threw that out.

nandedamana:
On 20/08/24 18:31, Kolja wrote:
>
> #3265 <https://github.com/flathub-infra/website/pull/3265> is arguably 
> even worse
>
Could you please elaborate? Asking only because I'm genuinely curious. I 
see a lot of cautionary comments related to this issue in general 
mentioning security, but it looks like the exact concerns are mentioned 
nowhere.

[...]

PR #3265, opened 2024-05-24
fix issue #3259: App verification step has incorrect domain computation

nandedamana:
- I have not tested this locally; too much work to set up everything just for a couple of line; sorry.
- I'm not satisfied with this fix; it won't handle sites with `www` prefix; but I think that was already broken

### DIFF
--- a/backend/app/verification.py
+++ b/backend/app/verification.py
@@ -135,8 +135,14 @@ def _get_domain_name(app_id: str) -> str | None:
         # https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/
         return f"{projectname}.{domain}.io".lower()
     else:
-        [tld, domain] = app_id.split(".")[0:2]
-        domain = _demangle_name(domain)
+        compons = app_id.split(".")
+        compons.pop(-1) # remove the app name
+
+        domain = _demangle_name(compons.pop(-1))
+
+        compons.reverse()
+        tld = ".".join(compons)
+
         return f"{domain}.{tld}".lower()
 
 


razzeee:
Duplicates https://github.com/flathub-infra/website/pull/2301/files

nandedamana:
> Duplicates https://github.com/flathub-infra/website/pull/2301/files

Sorry that I didn't notice that (had performed a quick search though). So what do you suggest in my case? Can I get the verification done manually? Whome should I contact?

backend/app/verification.py
```diff
@@ -135,8 +135,14 @@ def _get_domain_name(app_id: str) -> str | None:
         # https://sourceforge.net/p/forge/documentation/Project%20Web%20Services/
         return f"{projectname}.{domain}.io".lower()
     else:
-        [tld, domain] = app_id.split(".")[0:2]
-        domain = _demangle_name(domain)
+        compons = app_id.split(".")
+        compons.pop(-1) # remove the app name
+
+        domain = _demangle_name(compons.pop(-1))
```
bbhtt, line 141: pop removes the element then returns that value, so domain is being set to the last component which is incorrect.

For something like `in.co.foo.bar` you are calculating the website to be `bar.foo.co.in` when it should be `foo.co.in`

bbhtt: `return ".".join(reversed([ _demangle_name(i) for i in appid.split(".")[:-1]])).lower()`

nandedamana:
On 20/08/24 21:04, bbhtt wrote:
>
> For something like |in.co.foo.bar| you are calculating the website to 
> be |bar.foo.co.in| when it should be |foo.co.in|
>
I disagree. Think about people having websites on managed services like 
Vercel, which is a very common case. How can (and why should) the owner 
of gimp.vercel.app put a text file under vercel.app?

bbhtt:
We don't allow app ids with vercel or things like that 

You either take everything except the last component and reverse it or this is pretty useless. That's the general convention on which apps are allowed 

Taking all components makes zero sense

bbhtt:
The requirements are written in https://docs.flathub.org/docs/for-app-authors/requirements#application-id

The general rule is submitter or upstream has control over reversed appid[:-1]

foo.vercel.app won't be allowed because they don't own vercel.app

nandedamana:
On 20/08/24 21:47, bbhtt wrote:
>
> foo.vercel.app won't be allowed because they don't own vercel.app
>
Sorry if my previous comment was ambiguous. By "gimp.vercel.app", I mean 
the website "gimp.vercel.app". So the app ID would be 
"app.vercel.gimp.GIMP" (in the pattern of org.gimp.GIMP).

Based on the code you were suggesting, the author should put the text 
file under vercel.app, which they cannot, and based on what my code was 
trying to do, the author should put the text file under gimp.vercel.app, 
which they can.

My question is simple: if someone is allowed to have an app id 
TLD.DOMAIN.SUBDOMAIN.APPNAME, why ask them to put the verification file 
under DOMAIN.TLD? Why not ask them to put it under SUBDOMAIN.DOMAIN.TLD? 
Isn't this an extremely simple and effective solution?

bbhtt:
> Why not ask them to put it under SUBDOMAIN.DOMAIN.TLD?

That's exactly what the suggestion I gave is doing while yours is asking them to put it under `APPNAME.SUBDOMAIN.DOMAIN.TLD`

nandedamana:
On 20/08/24 22:06, bbhtt wrote:
>
> That's exactly what the suggestion I gave is doing while yours is 
> asking them to put it under |APPNAME.SUBDOMAIN.DOMAIN.TLD|
>
Are you sure about this claim? When I separate my code into a test file 
and call `print(getsite("TLD.DOMAIN.SUBDOMAIN.APPNAME"))`, I'm getting 
`subdomain.domain.tld` as the output.

My code was originally meant to handle cases like ".co.in", but if it 
works for subdomains, then the additional branches for github.io and the 
like can also be removed.

bbhtt:
> Are you sure about this claim? 

Yes see the review comment.

when you do `_demangle_name(compons.pop(-1))` did you expect to act it on a list? Because it can't and will show an exception. It's acting on the return value of pop() which later is going into `return f"{domain}.{tld}".lower()

bbhtt:
what you inputted as tld variable is also quite wrong. Actually you are using `tld = SUBDOMAIN.DOMAIN.TLD`

nandedamana:
On 20/08/24 22:26, bbhtt wrote:
>
> when you do |_demangle_name(compons.pop(-1))| did you expect to act it 
> on a list?
>
Of course not.

nandedamana:
> what you used as the `tld` variable is also quite wrong. You are using `tld = SUBDOMAIN.DOMAIN.TLD`

No. tld will be `DOMAIN.TLD` (which is still not a TLD, but that's okay since such a separation is meaningless).

Here's the output from a test script that uses the same code with some print statements:

```
app_id =  TLD.DOMAIN.SUBDOMAIN.APPNAME
domain =  SUBDOMAIN
tld =  DOMAIN.TLD
final =  subdomain.domain.tld
```

And here's the code:

```
def _demangle_name(domain):
    # Returning verbatim for testing
    return domain

def getsite(app_id):
    compons = app_id.split(".")
    compons.pop(-1) # remove the app name

    domain = _demangle_name(compons.pop(-1))
    print("domain = ", domain)

    compons.reverse()
    tld = ".".join(compons)
    print("tld = ", tld)

    return f"{domain}.{tld}".lower()

app_id = "TLD.DOMAIN.SUBDOMAIN.APPNAME"
print("app_id = ", app_id)
print("final = ", getsite(app_id))
```

Please note that I'm not saying the code is perfect. For instance, it only de-mangles the first component of a domain name. However, I'm puzzled by your claims as to what my code does. Maybe I'm missing something else here. I just want to know.

bbhtt:
Ok looks like I was dreaming, I missed that you did pop twice. The first one removes the appname part the second one removes subdomain and sets domain to the demangled subdomain.

However now the issue is you aren't demangling the TLD part. In this case your TLD is `domain.tld` which needs to be demangled as well.

nandedamana:
On 21/08/24 07:41, bbhtt wrote:
>
> Ok looks like I was dreaming, I missed that you did pop twice. The 
> first one removes the appname part the second one removes subdomain 
> and sets domain to the demangled subdomain.
>
Yes, exactly. Thanks for summing it up.
>
> However now the issue is you aren't demangling the TLD part. In this 
> case your TLD is |domain.tld| which needs to be demangled as well.
>
Yes, an issue I've acknowledged already. The reason I didn't mess with 
it is I don't know which parts are mangled. You should be familiar with 
the process. Please make the relevant changes yourself.

Also, since this solution originally meant for cc-TLDs seem to support 
subdomains as well, I think the nearby branches for specific cases like 
*.github.io can be removed, right?

bbhtt:
> The reason I didn't mess with
it is I don't know which parts are mangled.

In `DOMAIN.TLD` `DOMAIN` will always be mangled, you don't need know. If it's not mangled, then it violated the App ID rules.

> I think the nearby branches for specific cases like
*.github.io can be removed, right?

No because code hosting static pages doesn't support arbitrarily many subdomains you can't have `foo.bar.github.io`, those need to take exactly the first 3 components always, except for Gitlab.

bbhtt:
Or rather I should say if it was supposed to be mangled and it's not then it violated App ID rules.

Issue #114, opened 2022-04-08
Show OARS information


JakobDev:
The OARS informations are currently missing on the Website

razzeee:
I've been playing around with this for some days, trying to use the c api via pygobject to construct a content rating and call the minimum age method. But it seems like python still hates me.

https://github.com/ximion/appstream/blob/ad04eb71fa85b6c2fb5c62948cf01320407dc0ac/qt/contentrating.cpp#L109

razzeee:
When we do this, we should also expose it via schema.org if possible https://schema.org/contentRating

xi:
I am currently looking into this. I want to document my progress so far:

The code that was used in gnome-software was upstreamed in https://github.com/ximion/appstream/commit/18c8b18d5a131669fa25a01d283a06f896ca438c

We probably want to do something like this:

```python
import gi

gi.require_version("AppStream", "1.0")
from gi.repository import AppStream

LEVELS = {
    'unknown': AppStream.ContentRatingValue.UNKNOWN,
    'none': AppStream.ContentRatingValue.NONE,
    'mild': AppStream.ContentRatingValue.MILD,
    'moderate': AppStream.ContentRatingValue.MODERATE,
    'intense': AppStream.ContentRatingValue.INTENSE,
}


def get_content_rating_summary(data: dict, locale: str) -> tuple[str, str]:
    system = AppStream.ContentRatingSystem.from_locale(locale)

    age = max([
        AppStream.ContentRating.attribute_to_csm_age(key, LEVELS[value])
        for key, value in data.items()
    ], default=0)

    return (
        AppStream.ContentRatingSystem.format_age(system, age),
        AppStream.ContentRatingSystem.to_string(system),
    )


print(get_content_rating_summary({
    'violence-cartoon': 'moderate',
    'violence-fantasy': 'intense',
}, 'de_DE'))
# ('6', 'USK')
```

sidt4:
I think this issue should be addressed as a priority. Refer https://github.com/flathub/net.krafting.PleasureDVR/issues/2.

PR #3637, opened 2024-07-24
Draft: Content rating

xi:
See #114

I worked on this during GUADEC, and I am not sure if I will be able to continue working on this any further. Unfortunately I did not get it to a usable state, but I will try to note done all interesting details so someone else can pick this up and finish the remaining bits and pieces.

gnome-software worked on this a while ago and upstreamed its logic into libappstream (see


backend/app/utils.py
```diff
@@ -394,6 +397,33 @@ def appstream2dict(appstream_url=None) -> tuple[dict[str, dict], dict[str, dict]
...
+    rating = AppStream.ContentRating()
+    rating.set_kind(content_rating["kind"])
+    details = {"attrs": []}
+
+    for attr, level in content_rating.items():
+        if attr == 'kind':
+            continue
+        c_level = AppStream.ContentRatingValue.from_string(level)
+        rating.add_attribute(attr, c_level)
+        description = AppStream.ContentRating.attribute_get_description(attr, c_level)
```
xi, line 411: I am not sure where this gets its locale from.

backend/app/utils.py
```diff
@@ -394,6 +397,33 @@ def appstream2dict(appstream_url=None) -> tuple[dict[str, dict], dict[str, dict]
     return apps, apps_locale
 
 
+def get_content_rating_details(content_rating: dict, locale: str) -> dict:
+    system = AppStream.ContentRatingSystem.from_locale(locale)
+    rating = AppStream.ContentRating()
+    rating.set_kind(content_rating["kind"])
+    details = {"attrs": []}
```
razzeee, line 404: Pylance complains here, I guess `details = {}` should work fine here

backend/app/utils.py
```diff
@@ -394,6 +397,33 @@ def appstream2dict(appstream_url=None) -> tuple[dict[str, dict], dict[str, dict]
     return apps, apps_locale
 
 
+def get_content_rating_details(content_rating: dict, locale: str) -> dict:
```
razzeee, line 400: Did you work this into the update routine at some point already?

razzeee:
Looks promising, work should focus on the backend part for now

backend/app/utils.py
xi: No, because I wasn't sure how to. See my comment above:

> libappstream also provides localized descriptions of all content ratings. I was not sure how to integrate this with flathub's approach to i18n. We could either try to hook into that. Or we could duplicate the code.

razzeee:
`kind` seems to be wrong and should be `type`

razzeee:
Pushed something that has started to work, more problems are:

- description doesn't get translated, do we need to do that on our end?
- we set a locale which is for e.g. `de`, but that's not mapping clearly to a rating system so we could assume `de_DE`, but it might be wrong

PR #3848, opened 2024-09-11
Replace SMEMBERS usage with postgres

barthalion:


backend/app/summary.py
```diff
@@ -176,7 +176,12 @@ def parse_summary(summary, sqldb):
 
 
 def update(sqldb) -> None:
-    current_apps = {app[5:] for app in db.redis_conn.smembers("apps:index")}
+    current_apps = set(
+        app.app_id
+        for app in sqldb.session.query(models.Apps.app_id)
+        .filter(models.Apps.type.in_(["desktop", "console-application"]))
```
razzeee, line 182: wondering, if this might cause problems with addons

PR #3888, opened 2024-09-21
Remove locales, before adding new ones

razzeee:
We've kept translations around forever for each app, this causes issues.


razzeee:
This fails locally for me, but I'm unsure why. The error suggests, that redis-py is broken.

razzeee:
Seems to work now, but takes forever to complete. We should probably move this out of redis.

PR #4039, opened 2024-10-23
Translate setup pages

razzeee:
@AsciiWolf we might want to inverse the source system for the setup instructions when this is merged


AsciiWolf:
> @AsciiWolf we might want to inverse the source system for the setup instructions when this is merged

That would be great, sadly I am just a volunteer and my time is very limited. :-/ But feel free to send a PR, I would be happy to merge it!

PR #4068, opened 2024-10-30
Lowercase currencies when parsing

razzeee:


razzeee:
We should add a story for the history page elements

Issue #4105, opened 2024-11-05
Do not display screenshots with environment="windows"


CarlSchwan:
For example: https://flathub.org/apps/org.kde.neochat the 2 last screenshots should not be displayed because they are marked as for Windows https://invent.kde.org/network/neochat/-/blob/master/org.kde.neochat.appdata.xml?ref_type=heads#L402

bbhtt:
`environment="windows"` does not seem valid according to spec.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

> Optionally, a screenshot may also have an environment property. This string property denotes the GUI environment the screenshot was recorded in, in the form of {env}:{style}, where {env} is a desktop-environment name in lowercase and {style} is a specific style that the desktop environment recognizes, e.g. light and dark for light and dark themes.

It's not in the form of `env:style`, and Windows is not a DE.

razzeee:
But it's listed in https://github.com/ximion/appstream/blob/main/data/desktop-style-ids.txt

bbhtt:
Yea it's weird, @ximion are env's supposed to strictly follow the format or not?

CarlSchwan:
Also the spec is a bit wrong about `env:style`, for example only `gnome` and `gnome:dark` but not `gnome:light` are allowed because only values in the desktop-style-ids.txt are allowed.

bbhtt:
> Yea it's weird, @ximion are env's supposed to strictly follow the format or not?

Ah I guess this the exact same question as https://github.com/ximion/appstream/issues/667

PR #4106, opened 2024-11-05
utils: Ignore non-Linux screenshots

bbhtt:
Fixes https://github.com/flathub-infra/website/issues/4105


bbhtt:
greenlet does not build with python 3.13

bbhtt:
> I would prefer to not have `is_linux` exposed in the end, as it doesn't seem of much value?

Hm maybe I can expose `environment` instead directly.

razzeee:
Should be good to go, optional things I can think of:

1. Omit null, when not set (save data on api requests)
2.  Have a test, that consumes macos/windows screenshots and check, if they are removed

razzeee:
Thank you

PR #4117, opened 2024-11-08
Filter screenshots by environments on a higher level

razzeee:
Makes sure, we don't add translations for screenshots we don't have

@bbhtt

### DIFF
--- a/backend/app/utils.py
+++ b/backend/app/utils.py
@@ -108,8 +108,18 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
                         "".join(description),
                     )
 
-        screenshots = component.find("screenshots")
-        if screenshots is not None:
+        screenshotsComponent = component.find("screenshots")
+
+        if screenshotsComponent is not None:
+            screenshots = [
+                screenshot
+                for screenshot in screenshotsComponent
+                if screenshot.attrib.get("environment")
+                not in (
+                    "windows",
+                    "macos",
+                )
+            ]
             app["screenshots"] = []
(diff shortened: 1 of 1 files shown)


backend/app/utils.py
```diff
@@ -108,8 +108,18 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
...
+        if screenshotsComponent is not None:
+            screenshots = [
+                screenshot
+                for screenshot in screenshotsComponent
+                if screenshot.attrib.get("environment")
+                not in (
+                    "windows",
+                    "macos",
+                )
+            ]
```
bbhtt, line 122: I guess this is fine but it removes the "fallback" behaviour I added with all()

bbhtt, line 122:
```suggestion
        if screenshotsComponent is not None:
            if not all(
                [
                    screenshot.attrib.get("environment")
                    in (
                        "windows",
                        "macos",
                    )
                    for screenshot in screenshotsComponent
                ]
            ):
                screenshots = [
                    screenshot
                    for screenshot in screenshotsComponent
                    if screenshot.attrib.get("environment")
                    not in (
                        "windows",
                        "macos",
                    )
                ]
            else:
                screenshots = [screenshot for screenshot in screenshotsComponent]
```

razzeee: Should we have a test case if this is important?

bbhtt: I don't mind either way with tests.

I had the fallback bevause I don't want the backend to end uo with no screenshots just because someone incorrectly tagged them. We use what we have.

PR #4120, opened 2024-11-08
Fallback for no screenshots, if there are only windows or macos scree…

razzeee:
…nshots

Courtesy of @bbhtt


bbhtt:
Seems fine, translated pages work. Thanks.

PR #4159, opened 2024-11-18
Align setup according to https://github.com/leerob/next-self-host/blob/main/deploy.sh#L139

razzeee:


nginx/nginx.conf
```diff
@@ -108,6 +108,15 @@ http {
 
         location / {
             proxy_pass http://frontend:3000/;
+            proxy_http_version 1.1;
```
razzeee, line 111: @barthalion I'm not too sure if we want these changes

razzeee: Yes, not sure if there's any harm in having it anyway, but I guess using nginx defaults would be fine - even if they might be very conservative.

PR #4488, opened 2025-01-17
utils: Fix isMobileFriendly calculation from display length 

bbhtt:


bbhtt:
Dropped the lower bounds of the ranges, seems like appstream can correctly catch invalid values like `-1` etc.

PR #4582, opened 2025-02-04
Fix rtl chevron in breadcrumbs

### LINKED ISSUE #4576: incorrect arrow direction in RTL
At https://flathub.org/ar/apps/category/game/1 for example (this is the case with any category/directory page)

Image

Arrow should be facing the other direction

razzeee:
Closes #4576


jermanuts:
Thanks for the PR @razzeee but the spacing is not equal, at least for the first path

RTL

image

LTR

image

razzeee:
should improve with the next deploy

PR #4596, opened 2025-02-05
Set new submission and direct upload only if first_seen_at was absent

bbhtt:
Fixes 477ea9080e0e1e73c195e78f29c710acf4d01209


razzeee:
I was wondering what those notifications were about

PR #4612, opened 2025-02-08
Initial swap from redis to postgres

barthalion:


backend/app/summary.py
```diff
@@ -253,6 +251,39 @@ def update(sqldb) -> None:
...
+        except Exception as e:
+            print(f"Error encoding summary data for {app_id}: {str(e)}")
+            continue
+
+    # update all apps in a single transaction
+    try:
+        for app_id, summary_json in apps_to_update.items():
+            app = models.App.by_appid(sqldb, app_id)
+            if app:
+                appstream_data = app.appstream
```
razzeee, line 269: Do we really need to write the appstream data here too? It should not change here, correct? Can't we just touch summaries and be done?

backend/app/summary.py
```diff
@@ -253,6 +251,39 @@ def update(sqldb) -> None:
...
+                app.summary = summary_json
+                if appstream_data:
+                    app.appstream = appstream_data
+                sqldb.session.add(app)
+            else:
+                app = models.App(
+                    app_id=app_id,
+                    type="desktop-application",
```
razzeee, line 277: Could this cause problems/be wrong?

`summary.update()` also always runs after `load_appstream`, so it could be fine, to not have the else at all here?

backend/app/app.py
```diff
@@ -416,20 +416,42 @@ def get_summary(
     ),
     branch: Optional[str] = None,
 ):
+    with get_db("replica") as db_session:
+        app = models.App.by_appid(db_session, app_id)
+        if app and app.summary:
+            value = app.summary
```
razzeee, line 422: can we name this `summary` instead of value?

razzeee:
Some small nitpicks, seems fine otherwise

bbhtt:
Code seems fine to me.

backend/app/worker.py
```diff
@@ -87,10 +87,14 @@ def update():
...
+            with WorkerDB() as sqldb:
+                app = models.App.by_appid(sqldb, app_id)
+                if app and app.summary and "timestamp" in app.summary:
+                    created_at = app.summary["timestamp"]
+                elif metadata := db.get_json_key(f"summary:{app_id}:stable"):
+                    created_at = metadata.get("timestamp")
+                else:
+                    created_at = int(datetime.utcnow().timestamp())
 
             apps_created_at[app_id] = float(created_at)
```
bbhtt, line 99: it's getting set to None somehow

```
worker-1            | Traceback (most recent call last):
worker-1            |   File "/venv/lib/python3.12/site-packages/dramatiq/worker.py", line 487, in process_message
worker-1            |     res = actor(*message.args, **message.kwargs)
worker-1            |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
worker-1            |   File "/venv/lib/python3.12/site-packages/dramatiq/actor.py", line 185, in __call__
worker-1            |     return self.fn(*args, **kwargs)
worker-1            |            ^^^^^^^^^^^^^^^^^^^^^^^^
worker-1            |   File "/app/app/worker.py", line 99, in update
worker-1            |     apps_created_at[app_id] = float(created_at)
worker-1            |                               ^^^^^^^^^^^^^^^^^
worker-1            | TypeError: float() argument must be a string or a real number, not 'NoneType'
```

backend/app/worker.py
```diff
@@ -87,10 +87,14 @@ def update():
...
-            if metadata := db.get_json_key(f"summary:{app_id}:stable"):
-                created_at = metadata.get("timestamp")
-            else:
-                created_at = int(datetime.utcnow().timestamp())
+            with WorkerDB() as sqldb:
+                app = models.App.by_appid(sqldb, app_id)
+                if app and app.summary and "timestamp" in app.summary:
+                    created_at = app.summary["timestamp"]
+                elif metadata := db.get_json_key(f"summary:{app_id}:stable"):
+                    created_at = metadata.get("timestamp")
```
bbhtt, line 95: metadata is not supposed to have a timestamp?

backend/app/summary.py
razzeee: Hopefully the type would be corrected by the next run of whatever, but it might be wrong on creation. Not sure, if the other logic will correct type. Would need to check.

PR #4642, opened 2025-02-11
Update distro.yml with corrections for Solus

infinitymdm:
Strictly speaking, a reboot isn't necessary either, but leaving it in since it can't really hurt.

### DIFF
--- a/frontend/public/locales/en/distros.json
+++ b/frontend/public/locales/en/distros.json
@@ -192,6 +192,7 @@
   },
   "solus": {
     "distroName": "Solus",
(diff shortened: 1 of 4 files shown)


razzeee:
https://github.com/flatpak/flatpak.github.io/blob/source/data/distro.yml should be kept in sync

infinitymdm:
> https://github.com/flatpak/flatpak.github.io/blob/source/data/distro.yml should be kept in sync

Which of these is the "source of truth" for this file? Should I make a PR there and refer to this one?

razzeee:
We need to sync them by hand at this point. Originally the linked file was the source of truth, but as we have the more complicated setup with translations etc, we should probably invert it now.

PR #4652, opened 2025-02-12
Another round of routes refactor

barthalion:


razzeee:
Looks good so far, will need to regenerate orval.

But we might also want to change the tags for some of these? Not sure.

Issue #4137, opened 2024-11-12
[Distribution Pages] Setup Instructions Updates


RacerBG:
**Ubuntu**

Ubuntu 18.04 is EOL since May 2023 and only ESM (paid) support is available since then. This is not something which the average user will do and they should probably be using 20.04 LTS or 22.04 LTS by now. The instructions for 18.04 can safely be removed from the setup instructions to simplify the translations and to eliminate the redundant information.

Relevant text:

```
With older Ubuntu versions, the official Flatpak PPA is the recommended way to install Flatpak. To install it, run the following in a terminal:
 sudo add-apt-repository ppa:flatpak/stable
 sudo apt update
 sudo apt install flatpak
```

_Source_: https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices

**ChromeOS**

The following statement is not wrong but it highly depends on which version of ChromeOS is installed on the Chromebook because only Debian Bullseye have available backport for flatpak: `A more up to date flatpak package is available in the [Debian backports repository](https://backports.debian.org/Instructions/).`

_Source_: https://packages.debian.org/search?searchon=names&keywords=flatpak

**openSUSE**

The following statement needs better wording: `If your distribution version is not shown by default, click Show flatpak for other distributions first and then select from the list.` For example: `If your distribution version is not shown by default, click Show under Unsupported distributions and then select from the list.`

_Source_: https://software.opensuse.org/package/flatpak

Pop!_OS

The same situation as Ubuntu.

Relevant text:

```
If you are using an older version, you can refer to the instructions below.
Install Flatpak

To install Flatpak on Pop!_OS 19.10 and earlier, open the Terminal app and run:
sudo apt install flatpak
Add the Flathub repository

Flathub is the best place to get Flatpak apps. To enable it, download and install the [Flathub repository file](https://dl.flathub.org/repo/flathub.flatpakrepo) or run the following in a terminal:
flatpak remote-add --user --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
Restart

To complete setup, restart your system. Now all you have to do is [install apps](https://flathub.org/)!

Note: graphical installation of Flatpak apps may not be possible with Pop!_OS 19.10 and earlier.
```

**Raspberry Pi OS**

[...]

PR #4661, opened 2025-02-15
Setup Instructions Updates

RacerBG:
Fixes: #4137 

Also it should unblock the following PR: https://github.com/flathub-infra/documentation/pull/391

Regards,
Georgi


razzeee:
https://github.com/flatpak/flatpak.github.io/blob/source/data/distro.yml should be updated in sync

RacerBG:
> https://github.com/flatpak/flatpak.github.io/blob/source/data/distro.yml should be updated in sync

Will do.

PR #4667, opened 2025-02-15
Type search endpoints

razzeee:


razzeee:
@barthalion working on this, made it seem, like we should be using `model_validate` in way more places

PR #4702, opened 2025-02-22
Process moderation requests in a background task

barthalion:


bbhtt:
seems fine.

razzeee:
did the need for this arise due to something or was this just natural cleanup/refactoring?

PR #4703, opened 2025-02-23
Make flathub github default bugtracker and override it for unverified

bbhtt:


razzeee:
Would be cool if we could find a way to not break these on platform e.g. https://flathub.org/apps/org.gnome.Platform

bbhtt:
@barthalion can verify them manually, runtimes are manually verified.

bbhtt:
Looks like they are already verified but these have leftover non EOL branches in catalogue which do not have the verification tag.

bbhtt:
Nevermind, it is doing the right thing, none of the runtimes are setting a bugtracker URL so it is correct to override them to Flathub.

https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/efd9e517cd82d90464a48065dab72b668c2dbfe7/files/os-release/org.gnome.Platform.appdata.xml.in#L30

https://invent.kde.org/packaging/flatpak-kde-runtime/-/blob/c64a56d117874379357eb522f523155c184e588f/org.kde.Platform.metainfo.xml#L9

bbhtt:
Will add them to the runtimes.

PR #4704, opened 2025-02-23
Routes refactor: categories, keyword, stats

barthalion:


razzeee:
I don't think the frontend can build, without the backend deployed. So the pipeline likely won't be able to turn green.

I'll just merge and keep watching the service. It works fine locally.

PR #4734, opened 2025-03-01
Front-end: wording improvements

Tiagoquix:
- Adds "for" to the sentence in English.
- Update and improve the Portuguese translations.


razzeee:
This is supposed to match with https://gitlab.gnome.org/GNOME/gnome-software/-/blob/main/src/gs-safety-context-dialog.c#L235

Also, if you edit translations, only ever touch the `en` folder, everything else is off limits in the repo.

PR #4735, opened 2025-03-01
Enable httpx sentry integration

razzeee:


razzeee:
@barthalion I think it still makes sense to have this extra setup

PR #4736, opened 2025-03-01
Type collection routes completly and remove frontend meilisearch types

razzeee:


backend/app/search.py
```diff
@@ -547,7 +561,14 @@ def get_runtime_list():
     )["facetDistribution"]["runtime"]
 
 
-def get_developers(page: int | None, hits_per_page: int | None):
+class DevelopersResponse(BaseModel):
```
razzeee, line 564: @barthalion  `/developers` seems to be completely unused, should we just remove it?

PR #4758, opened 2025-03-05
Filter out EOL refs where we don't need them

barthalion:


backend/app/worker/update.py
```diff
@@ -13,7 +13,7 @@ def update():
         summary.update(db)
     exceptions.update()
 
-    current_apps = apps.get_appids()
+    current_apps = apps.get_appids(include_eol=True)
```
razzeee, line 16: Should not need it and IMO might be bad, as it can cause a meilisearch entry due to this

backend/app/summary.py
```diff
@@ -267,7 +267,7 @@ def fetch_summary_bytes(url: str) -> bytes | None:
 
 
 def update(sqldb) -> None:
-    current_apps = set(apps.get_appids())
+    current_apps = set(apps.get_appids(include_eol=True))
```
razzeee, line 270: Should not need it and IMO might be bad, as it can cause a meilisearch entry due to this

razzeee:
Ideally, we would have an eoled app in the smoketest suites summary file, so that we can check, that it doesn't surface.

backend/app/worker/update.py
razzeee: You probably wanted to comment this on `apps.py`. There it might make sense, at least we're skipping things that already don't exist, so it shouldn't hurt.

Then again, with the new context a line like

```
models.App.delete_app(sqldb, app_id)
```

doesn't seem correct anymore.

which in turn makes me wonder, if 

```
app_stats_key = f"app_stats:{app_id}"
if database.redis_conn.exists(app_stats_key):
            continue
```

broke

Issue #320, opened 2022-05-30
Adjust for population in per-country downloads


wjt:
The *Downloads per Country* map on https://beta.flathub.org/statistics is great data to publish. However, absolute download counts per country are IMO only useful in comparison to each country's population.

The World Bank provides [per-country population data](https://data.worldbank.org/indicator/SP.POP.TOTL).  I think it would be interesting to visualise downloads per 100,000 people (pick your own scale factor). Just picking a few:

| Country | Downloads | Downloads per 100,000 population |
| --- | --- | --- |
| US | 66 695 282 | 20 119 |
| RU | 28 447 666 | 19 741 |
| BR | 62 869 311 | 29 577 |

From this we can see a slightly different picture emerging: while the absolute numbers would suggest substantially more Flathub users in US than RU, and slightly more than BR, in fact US and RU are pretty much the same per capita and Brazil has 50% more Flathub downloads per capita.

razzeee:
I briefly looked at this and finding a source for the population data, that is easily updated and includes almost all countries was challenging. Someone else might have more luck, but this seems to add some maintenance burden.

pauladam94:
I think this can be reopened as this PR #5197 tries to respond to this issue.

PR #5197, opened 2025-06-16
Downloads Statistics per Population Country

pauladam94:
This PR respond to #320. Right now this PR should not be merged and just aggregate what has been said on the issue(#320) and [the flathub forum](https://discourse.flathub.org/t/different-statistics-that-raw-downloads-on-flathub-statistics-page/9820/7).

## This PR has two files
- a markdown file that explain how to fetch the data from a github repository that have population data by country regularly updated
- the translation data between iso2 and iso3 country names in a csv form

Ideally we should find a library that gives those directly from typescript.


razzeee:
> Ideally we should find a library that gives those directly from typescript.

Ideally, we handle that in python and offer it via the api to the frontend/typescript part

razzeee:
https://github.com/flathub-infra/website/blob/main/backend/app/stats.py#L164

Is what you're looking for. Not sure if we still want to expose the old/total count.
You probably don't need to interface with the database, you should be able to map it on the fly.

It gets returned here https://flathub.org/api/v2/stats/

razzeee:
I guess you could just generate a json file and add it to git and then consume that. 

But provide a script to update the json file.

pauladam94:
You prefer that than fetching data on the fly ?
You are thinking of a script running on CI maybe ?
I do agree this might be a better approach for the backend to never fail when running. The only potential break will be on CI when fetching data which is not a problem because it will not crash the backend.


Where can I look for the typescript code that is getting the data ? Because there will be a small breaking change of the final json. Also where is the map generated ?

pauladam94:
Do you have a preferred place for the place of this script (and name) ? Same for the place of the data generated.

The script will create a csv (or json. You prefer json ?) with this data :

```
country_iso3, population
us, 100
fr, 200
...
```

pauladam94:
What does this line means ?
https://github.com/flathub-infra/website/blob/08a0ee450ac86566c2b850eb8a153997772b9ee0/frontend/pages/statistics.tsx#L140

value is the number of downloads for a the current country_code, what does that mean to translate this number ?
I assume this is some kind of translation because of i18 mention.

pauladam94:
I think I have done most of the work. The script you talked about is well documented, I think. I run the last version of it and it works. It has one dependency, the `requests` package for doing the http request. 

Do you want me to put leave the `requirements.txt` file ?
I also left (for now) my `shell.nix` file for this script (declarative dependency file for nix package manager) it can of course be removed.

How can I test the frontend code ? I mostly changed the value associated with the variable "country_code". I renamed the fiels "downloads_per_people".

In the current state of the PR, the statistics about the raw number of downloads is completely removed from the website (even if it can be queried because the backend sends it to the frontend).

pauladam94:
TLDR :
- [x] getting iso2 -> iso3 table
- [x] script for fetching iso3 -> population data
- [x] change number_dowloads -> dowloads_per_person in WorldMap in the frontend  
- [ ] test the backend 
- [ ] test the frontend

pauladam94:
I don't manage to run the docker of the backend, unfortunately. I get this error (I'm on NixOs) :
```
 => ERROR [worker stage-1 2/5] RUN apt-get update &&     apt-get install -y --no-install-recommends     python3 python-is-pyt  0.0s
 => [worker internal] load build context                                                                                       0.0s
 => => transferring context: 18.61kB                                                                                           0.0s
------
 > [worker stage-1 2/5] RUN apt-get update &&     apt-get install -y --no-install-recommends     python3 python-is-python3 python3-venv ca-certificates libcairo2 libpq5     libgirepository-2.0-0 gir1.2-ostree-1.0 gir1.2-appstream-1.0 &&     apt-get clean && rm -rf /var/lib/apt/lists/*:
------
failed to solve: process "/bin/sh -c apt-get update &&     apt-get install -y --no-install-recommends     python3 python-is-python3 python3-venv ca-certificates libcairo2 libpq5     libgirepository-2.0-0 gir1.2-ostree-1.0 gir1.2-appstream-1.0 &&     apt-get clean && rm -rf /var/lib/apt/lists/*" did not complete successfully: fork/exec /nix/store/kjg6q8g2fwjsnil88vi5v64qy166slbw-moby-27.5.1/libexec/docker/runc: exec format error
```

I don't know if it's worth it to try harder to fix my problem. My code changes are not that big, I think there are not going to break the existing backend and frontend.

backend/data/gen/population_country_fetch.py
```diff
@@ -0,0 +1,349 @@
...
+    'VIR' : 'VI',
+    'WLF' : 'WF',
+    'ESH' : 'EH',
+    'YEM' : 'YE',
+    'ZMB' : 'ZM',
+    'ZWE' : 'ZW',
+    'ALA' : 'AX',
+}
+
+tmpfile = tempfile.NamedTemporaryFile(delete=False, suffix='.zip')
```
razzeee: why is the suffix `zip`?

razzeee: NVM, the csv link downloads a zip

backend/data/gen/population_country_fetch.py
```diff
@@ -0,0 +1,349 @@
...
+# Checking command line arguments
+if len(sys.argv) <= 2:
+    print(f"Suppose to use it like that : python3 {sys.argv[0]} {{year}} {{output_file}}")
+    exit(-1)
+year = sys.argv[1]
+output_filename = sys.argv[2]
+
+'''
+    This script should be called like this :
+    > python3 population_count_fetch 2023 ../country_population.json
```
razzeee:
```suggestion
    > python3 population_country_fetch.py 2023 ../country_population.json
```

pauladam94:
I have fix the formatting of the frontend.

Here is the error on the backend : (I don't understand how my code has broken that)
```
Run astral-sh/setup-uv@v6
Trying to find required-version for uv in: /home/runner/work/website/website/uv.toml
Could not find file: /home/runner/work/website/website/uv.toml
Trying to find required-version for uv in: /home/runner/work/website/website/pyproject.toml
Could not find file: /home/runner/work/website/website/pyproject.toml
Could not determine uv version from uv.toml or pyproject.toml. Falling back to latest.
manifest-file not provided, reading from local file.
Error: ENOENT: no such file or directory, open 'version-manifest.json'
```

pauladam94:
Ok I will try again with another distro to run the backend, didn't understand why it crashes. Is there anything that runs on every python file in the repo ? It seems that the problem is because added a python file at a random place maybe.

razzeee:
Fix the data structure errors first, I suspect that that will fix it

PR #5213, opened 2025-06-21
Update localize.ts

koshan0othman:
wrong flag of ckb language.


koshan0othman (commented):
kurdish language doesn't have emoji flag, and iran flag isn't kurdish flag

razzeee:
> kurdish language doesn't have emoji flag, and iran flag isn't kurdish flag

I'm not following, we don't even have iran mentioned?

PR #5424, opened 2025-08-14
Add redirect for buildbot.flathub.org to pipelines

### LINKED ISSUE #5423: buildbot.flathub.org website times out
The sub-domain should either be phased out completely so the name can't be resolved, or the page should redirect to the vorarbeiter website.

razzeee:
Fixes #5423

Not sure if this will work

Waiting for @barthalion to merge

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -93,6 +93,12 @@ http {
     }
 
     server {
+        listen 80;
+        server_name buildbot.flathub.org;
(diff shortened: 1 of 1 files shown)


hadess:
Very useful, thanks!

Is it expected that it further redirects to https://flathub.org/builds ?

razzeee:
yes, that's what it does right now

PR #5496, opened 2025-08-28
Add quality guideline for end of life runtimes

razzeee:
It's automatically set


bbhtt:
zscore parts is missing, we probably want this to effect by a larger factor than others.

If not just go with half of the zscore logic I said, ignore this factor.

razzeee:
Yeah, this still affects zscore by default - not more then other failures. I think we should merge this first and then do the zscore changes.

PR #5504, opened 2025-08-30
Add new safety labels

bbhtt:


bbhtt:
> It could make sense to first land something like that and then rebase this upon that.

Yea that's fine.

razzeee:
https://github.com/flathub-infra/website/pull/5513 might be enough to give you a baseline

PR #5507, opened 2025-08-31
Halve base trending score if an app is eol

razzeee:


bbhtt:
Erm, we needed to check if runtime is EOL.

razzeee:
https://github.com/flathub-infra/website/pull/5514

PR #5535, opened 2025-09-09
Split search bar into component

razzeee:


razzeee:
Continued in https://github.com/flathub-infra/website/pull/5534

Issue #3131, opened 2024-05-05
Feedback on the unverified badge from upstream


barthalion:
> @barthalion the root issue isn't resolved I think.
> 
> To me, it is confusing that the above mentioned Flatpak site looks like the Genay Flatpak package is maintained by the Geany developers. This is because of the "by Geany Team" term below the title *and* in the blog "Community built" there is a link to the project website directly.
> 
> But at no point there is any reference to who really maintains this package and that bug reports related to the Flatpak packaging should be reported there and not to the Geany project itself.
> 
> I think it would help all parties, if this could be made more clear that project developers and package maintainers are not the same people.
> 
> Thank you.

_Originally posted by @eht16 in https://github.com/flathub/org.geany.Geany/issues/92#issuecomment-2094813672_

razzeee:
Duplicates/could be solved by https://github.com/flathub-infra/website/issues/3125

But that would be something @ximion would need to add to libappstream or we need to force people to login via github once.

TiZ-HugLife:
Hi there. As someone maintaining a package who is completely separate from the people actually making it--the referenced package, even--I would also consider myself a stakeholder here. Packager / developer disjoint is standard in traditional packaging, but given Flathub's huge push toward verification, I can see how the way it's currently presented on the site is problematic. I think that in unverified situations, it would be beneficial to either not show the author at all, or show both the upstream developer per the metadata and also show everyone who has worked on the package recently. Given a packager's role is comparatively small--especially in Flathub where it's kinda easy--I'm not sure I would want it to be *prominent,* but I would definitely want it to be *easily available.*

kekkoudesu:
Hi there! Since my issue was mentioned (and I'm also in the process of creating [my own Unverified Flatpak](https://github.com/kekkoudesu/lightworks-flatpak)), I thought I would chime in with some more thoughts. I'm glad this is being discussed.

Even at the time of creation, I didn't think #3125 was a great solution. The important information is not necessarily who is responsible for the package, but who *isn't*. That is, Google is *not* responsible for the Flatpak package.

So, you could remove the "by Google" line and add it to the Unverified badge:

![image](https://github.com/flathub-infra/website/assets/88186997/d343522a-b328-4811-ab9a-9b7b324da0a0)

It needs some more work (maybe split it into two badges?), but I like this direction more. It's simpler and only a small alteration to the way the page is presented, not requiring changes elsewhere.

[...]

PR #5725, opened 2025-10-01
Add more info to unverified badge

bbhtt:
Fixes https://github.com/flathub-infra/website/issues/3131


frontend/src/components/application/Description.tsx
```diff
@@ -19,13 +19,19 @@ export const Description = ({
   const description = useMemo(
     () =>
       app.description
-        ? linkifyHtml(sanitizeAppstreamDescription(app.description), {
-            rel: "noopener noreferrer",
-            target: "_blank",
-            validate: {
-              url: (value) => /^https:\/\//.test(value),
+        ? linkifyHtml(
+            sanitizeAppstreamDescription(app.description).replace(
```
razzeee, line 23: Should we maybe do this replace in the backend?

bbhtt: I personally prefer keeping stylisation to the frontend, so that the source of truth does not differ between the actual appstream data, and also it affects the API too. b is not a valid appstream tag.

razzeee: Fair if we expect the api to talk appstream and not html

PR #5730, opened 2025-10-02
Improve display of short changelogs

razzeee:


frontend/src/components/application/Releases.tsx
```diff
@@ -52,20 +52,22 @@ const Releases: FunctionComponent<Props> = ({
...
+  const { getCollapseProps, getToggleProps, isExpanded, setExpanded } =
+    useCollapse({
+      collapsedHeight: collapsedHeight,
+      defaultExpanded: expanded,
+    })
+
   const ref = useCallback((node) => {
     if (node !== null) {
       setShowCollapseButton(node.scrollHeight > collapsedHeight)
+      setExpanded(node.scrollHeight <= collapsedHeight || expanded)
```
bbhtt, line 67: This breaks expanding it manually, the callback logic should run only once i think.

razzeee: This only runs on init right now

PR #5904, opened 2025-11-12
utils: Filter and limit keywords

bbhtt:


backend/app/utils.py
```diff
@@ -410,6 +410,16 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
                         and tag.text is not None
                         and tag.text.strip()
                     ]
+
+                    if elem.tag == "keywords" and translated_items:
+                        translated_items = sorted(
```
razzeee: I don't like the idea of sorting, as I would hope, that the most relevant items are first in the keywords. So we would be loosing meaning here. Especially if we at some point decide, that every app only gets a max keyword amount.

backend/app/utils.py
```diff
@@ -410,6 +410,16 @@ def appstream2dict(appstream_url=None) -> dict[str, dict]:
...
+                            [
+                                item
+                                for item in translated_items
+                                if not any(c in item for c in "./")
```
razzeee: What's the `./` for and why is the other version below with a space in front?

bbhtt: Blocks words with ., / or space in them. Space is only applied to source locale (en).

backend/app/utils.py
bbhtt: Ok sorting can be dropped.

PR #5933, opened 2025-11-18
feat: add instant search

razzeee:


razzeee:
@barthalion can you land this? might be easier for you, with the env vars

PR #5947, opened 2025-11-22
Add keywords moderation

razzeee:


razzeee:
This will probably flag each app, that gets an update, as keywords are empty for all right now. So we'll need an answer to that.

bbhtt:
Then you don't know.

PR #5961, opened 2025-11-24
Add cache invalidation calls to update function

razzeee:


barthalion:
Use mark_stale instead of invalidate. Also https://github.com/flathub-infra/website/blob/0b79ab6829dae09d07b74f92741c6cde9326cec6/backend/app/update.py#L25 needs to be cleaned up if you want to keep it in the worker.

PR #6052, opened 2025-12-19
Type cleanup

razzeee:


backend/app/vending/__init__.py
```diff
@@ -478,10 +478,11 @@ def post_app_vending_setup(
         elif setup.recommended_donation > 0:
             if vend.user != login["user"].id:
                 raise VendingError(error="user-mismatch")
-            vend.currency = setup.currency
-            vend.appshare = setup.appshare
-            vend.recommended_donation = setup.recommended_donation
-            vend.minimum_payment = setup.minimum_payment
+            if vend:
```
barthalion: We are already accessing `vend` before the null check, the guard should be either elsewhere or `elif` implies vend is present

backend/app/models.py
```diff
@@ -377,17 +379,27 @@ def dev_flatpaks(self, db):
 
     def add_role(self, db, roleName: "RoleName") -> "FlathubUser":
         role = Role.by_name(db, roleName)
+        if role is None:
```
barthalion, line 382: This needs to be checked if returning ValueError won't explode elsewhere

razzeee: This should be fine, as it's part of `add_role`

backend/app/vending/__init__.py
razzeee: Refactored this

PR #6304, opened 2026-03-09
distro: Update Alpine Linux instructions, Add postmarketOS instructions

fossdd:
* Use canonical distribution name
* Drop gnome-software/discover plugin mentions
  (they are automatically installed with flatpak)


frontend/src/data/distro.yml
```diff
@@ -235,19 +235,13 @@
         <p>To complete setup, restart your system. Now all you have to do is <a href="https://flathub.org/">install apps</a>!</p>
         <p>Note: graphical installation of Flatpak apps is available only through GNOME Software and/or KDE Discover.</p>'
 
-- name: Alpine
+- name: Alpine Linux
```
razzeee: We probably don't want this, as this changes the url and would lead to 404 errors. (need to runtime check this, just a feeling)

fossdd: Ah, right. Didn't think of that. Will revert it back then, also not a big deal.

frontend/src/data/distro.yml
```diff
@@ -503,3 +497,9 @@
     <h2>Flatpak is installed by default on Salix since version 15.0—no setup required!</h2>
     <p>Flatpak comes preconfigured with the Flathub repository and desktop integration tools are included to allow 1-click install from Flathub.</p>
     <p>All you have to do is <a href="https://flathub.org/">install apps</a>!</p>
+
+- name: postmarketOS
```
razzeee, line 524: I'm not sure if we actually want to keep on adding these "non instructions".

They made sense, when distros first moved from not having flatpak to shipping with it.

razzeee:
This also needs changes in Distros.tsx

PR #6444, opened 2026-04-07
feat(frontend): Add HealthFitness category

radmorecameron:
HealthFitness was recently added as a main category on FreeDesktop xdg-specs.
https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/98#note_3339833
https://specifications.freedesktop.org/menu/latest/category-registry.html

This change should add support for displaying this category in Flathub.
I'm having some issues running the app locally with podman compose so I wasn't able to test this change sadly :frowning_face:

### DIFF
--- a/frontend/app/[locale]/(home)/page.tsx
+++ b/frontend/app/[locale]/(home)/page.tsx
@@ -36,6 +36,7 @@ const categoryOrder: MainCategory[] = [
   MainCategory.network,
   MainCategory.development,
   MainCategory.science,
+  MainCategory.healthfitness,
   MainCategory.system,
   MainCategory.utility,
 ]
--- a/frontend/app/[locale]/home-client.tsx
+++ b/frontend/app/[locale]/home-client.tsx
@@ -55,6 +55,10 @@ interface HomeClientProps {
 function MobileSection({ mobile }: { mobile: MeilisearchResponseAppsIndex }) {
   const t = useTranslations()
 
+  if (mobile.hits.length === 0) {
+    return null
+  }
+
   return (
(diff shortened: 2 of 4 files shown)


razzeee:
if merged like this, it wouldn't show anything - we need to hide it, when the section is empty. I can do that.

But we should start hinting at apps, that this new category exists.

@AlexanderVanhee will bazaar show apps that set this?

we will need to ask g-s the same thing https://gitlab.gnome.org/GNOME/gnome-software/-/work_items/2997

I've also reached out to KDE - https://invent.kde.org/plasma/plasma-workspace/-/work_items/185

AlexanderVanhee:
The fix to not follow the API is already out for a while now, so it should be fine. 

I would also be really suprised if GNOME Software depended on this.

razzeee:
@AlexanderVanhee sure, but will apps that set this new category still be discoverable?

razzeee:
@yakushabb I'll probably try to come up with a list of app candidates and will raise issues on their side. Just so you know.

AlexanderVanhee:
> @AlexanderVanhee sure, but will apps that set this new category still be discoverable?

I will try to add the new category as a search option, and even add it to the Explore page if you decide to add it to the Flathub website as well.

However, a new button on the Search page might look a bit weird, as it would start a new row. We might be able to replace the Adwaita category, though, but probably only when there are enough apps.

radmorecameron:
I believe https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/merge_requests/32 will need to be merged for apps to be able to use the new category

PR #6455, opened 2026-04-08
Revert "Revert "feat(backend): Add HealthFitness category""

razzeee:
Reverts flathub-infra/website#6454


radmorecameron:
Hmm, wasn't aware that the change would have broken Bazaar. It looks like a fix was merged and released in Bazaar: https://github.com/bazaar-org/bazaar/pull/1360 

Might be good to wait at least a month (maybe even longer) after the next Bazzite release to merge this so we don't break anything for them.

PR #6711, opened 2026-06-07
Add support for generating runtime tokens

barthalion:


backend/app/models.py
```diff
@@ -1537,6 +1537,25 @@ def delete_user(db, user: FlathubUser):
...
 
+class RuntimeScope(Base):
+    __tablename__ = "runtimescope"
+
+    id = mapped_column(Integer, primary_key=True)
+    app_id = mapped_column(String, nullable=False, unique=True, index=True)
+    prefixes = mapped_column(String, nullable=False)
+    extra_ids = mapped_column(String, nullable=False, server_default="")
+    repos = mapped_column(String, nullable=False, server_default="stable beta")
+    created_at = mapped_column(DateTime, nullable=False, server_default=func.now())
```
razzeee, line 1548: I would usually also add an updated_at

PR #6720, opened 2026-06-09
fix: mark current pagination button

remove334:
## Summary

- adds `type="button"` to callback-mode pagination buttons
- marks the current callback-mode page with `aria-current="page"`, matching the link-rendered pagination path
- adds a Storybook story for the `onClick` pagination variant

## Test plan

- `git diff --check`

### DIFF
--- a/frontend/src/components/Pagination.stories.tsx
+++ b/frontend/src/components/Pagination.stories.tsx
@@ -64,3 +64,13 @@ export const WithQueryParams: Story = {
     searchParams: mockSearchParams,
   },
 }
+
+export const WithOnClick: Story = {
+  args: {
+    currentPage: 3,
+    pages: [1, 2, 3, 4, 5],
+    onClick: () => undefined,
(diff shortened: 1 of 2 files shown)


razzeee:
Which page do you have this problem with?

remove334:
> 您在哪个页面上遇到这个问题？

This is about the callback/button variant of the shared Pagination component rather than a visual issue on a single page.

One current usage is `frontend/src/components/user/UserApps.tsx`, where the user apps list renders:

<Pagination currentPage={page} pages={pages} onClick={setPage} />

Another one is `frontend/src/components/application/AppVendingControls/OwnershipTokens/TokenList.tsx`.

In these cases Pagination renders page numbers as buttons. The link-based variant already sets `aria-current="page"` for the current page, but the button/onClick variant did not, so screen readers would not get the same current-page indication. This PR makes the button variant match the link variant and also adds `type="button"`.

PR #6730, opened 2026-06-10
Add archive button to direct uploads panel

barthalion:


bbhtt:
FYI, the frontend part of this PR was reverted by 619b3b890a73f2b025245fa79c27951e901c1f9a I don't think it was intentional as the UI part of archive / unarchive is now missing.

barthalion:
Restored. Dealing with sinuses infection and I'm half dead at times.

PR #6760, opened 2026-06-19
verification: Add an API to mark appid as manually verified

bbhtt:


barthalion:
Yep yep, makes sense to me. I really don't want to be a choking point on things, if there's anything else that comes to your mind, let me know.

BTW there's a workflow that will handle the codegen for you (and can also be triggered by hand if you push to main like I do sometimes)

bbhtt:
This is mainly because I don't like operating on the DB and openshift logs me out every day. I did it yesterday for torrentsdigger https://github.com/flathub/flathub/issues/9009 but it's annoying.

Now I can do it from the phone etc.

The other thing I want to look at is editing direct upload users for an apps / runtimes.

barthalion:
Editing who owns the app/is secondary maintainer is kinda iffy mostly because the original design assumes the primary maintainer invites any additional collaborators with his invite code (or something along this). But yeah, maybe we need general UI for modifying who is the primary maintainer and who can access the app additionally

barthalion:
Updated the existing view so it can change maintainers. Probably needs some tuning because UX is shit but always some starting point.

PR #6768, opened 2026-06-21
ci: Pin all actions and restrict permissions

bbhtt:


bbhtt:
Turned on the `Require actions to be pinned to a full-length commit SHA` setting, let me know if something breaks.

PR #6791, opened 2026-07-02
Update button label for managing applications

Justinzobel:


razzeee:
Not sure we want this. Would have to be translated at least.

It's also not working for admins, without access to all repos, so the condition is deciding on the wrong field.

bbhtt:
It's fine, but you should make it translatable.

Justinzobel:
> Not sure we want this.

It looks bad in my opinion because the two buttons either side of it have text.

> Would have to be translated at least.

What do I need to add to make it translatable?

> It's also not working for admins, without access to all repos, so the condition is deciding on the wrong field.

It worked for me on an app I have access on.

bbhtt:
You have to add a string to frontend/public/locales/en/common.json then `{t("manage")}`.

razzeee:
We should also find different icons for one or the other then. Having two buttons with the same icon is confusing / doesn't look good.

Justinzobel:
> We should also find different icons for one or the other then. Having two buttons with the same icon is confusing / doesn't look good.

I only see 3 buttons, and they all have unique icons:

image

Is there a button I don't see that has the cog icon as well? If so that is a separate issue.

razzeee:
If you are a moderator the details button becomes moderate with a cog icon

bbhtt:
Can you just add the label and make it translatable for now?

Justinzobel:
> Can you just add the label and make it translatable for now?

Done.

frontend/src/components/application/QualityModeration.tsx
```diff
@@ -293,7 +293,7 @@ export const QualityModeration = ({
           {canManageApp && (
             <Button size="lg" variant="secondary" asChild className="me-2">
               <Link href={`/apps/manage/${app.id}`}>
-                <Cog6ToothIcon className="size-5" />
+                <Cog6ToothIcon className="size-5" /> {t("manage")}
```
razzeee:
```suggestion
                <Cog6ToothIcon className="size-5" /> <div className="hidden sm:block">{t("manage")}</div>
```

PR #6959, opened 2026-08-15
Add curated selection layouts to backend

razzeee:
Persist grid and carousel choices for scheduled homepage selections and expose them through public and admin responses. Include fullscreen metadata with a bulk lookup so carousel banners can match weekly app positioning.

@AlexanderVanhee


AlexanderVanhee:
Looks good, how will the carousel show up in the UI? Will it have a title or subtitle aswell?

Kinda like you see here?

image

razzeee:
Haven't really decided on that, will explore that in the next step. Mainly wanted to make sure, that the api change doesn't break stuff for you.

AlexanderVanhee:
Adding keys should not break stuff for us.

PR #6980, opened 2026-08-23
Revert "Use text balance in some places"

razzeee:
This reverts commit 28ace5b85ad78eaee5ac84d2bc472ee08e247e93.


razzeee:
We might want to revisit this at some point when firefox supports `text-pretty`

